PR #65, opened 2021-02-16
NEXT-12712 - Add custom pagelet

Dominik28111:


guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,215 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @RouteScope(scopes={"storefront"})
+ */
+class ExampleController extends StorefrontController
+{
+    /**
+    * @Route("/example", name="frontend.example.example", methods={"GET"})
```
mitelg, line 65: the name of the route is important. there are two or three keywords a route name should start with. like frontend. otherwise features like SEO won't work

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,215 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @RouteScope(scopes={"storefront"})
+ */
+class ExampleController extends StorefrontController
+{
+    /**
+    * @RouteScope(scopes={"storefront", "custom-scope"})
```
mitelg, line 104: I won't bring the custom example here. otherwise you need to explain what's its purpose and that would be worth an own article :grin:

guides/plugins/plugins/storefront/add-custom-page.md
```diff
@@ -2,19 +2,25 @@
...
-## Adding custom storefront page
+In the following sections, we'll create each of the necessary classes one by one.
+The first one will be controller, whose creation is not going to explained here again. Have a look at the guide about [creating a controller](./add-custom-controller.md)
+to see why it works.
 
-### Storefront Controller Class example
+### Creating ExampleController
 
-First of all we have to create a new controller which extends from the `StorefrontController`. A controller is also just a service which can be registered via the DI-Container. Furthermore, we have to define our `RouteScope` via annotation, it is used to define which domain a route is part of and **needs to be set for every route**. In our case the scope is `storefront`. Now we can create an example function with a `Route` annotation which has to contain our route, in this case it's `/example`. Below you can find an example implementation where we render an `example.html.twig` template file with a template variable `example`.
+As already mentioned, the controller creation is not explained here, just an example controller.
```
mitelg, line 23: third time you mentioned that you are not going to explain it 😃

guides/plugins/plugins/storefront/add-custom-page.md
```diff
@@ -35,138 +39,257 @@ use Symfony\Component\Routing\Annotation\Route;
...
-Next, we need to register our controller in the DI-container and make it public.
+The first one is not necessary, but useful, since it loads all kind of default page stuff, such as a footer and a header and loads some additional
+helpful data. Once again, you don't have to do that, but if you want your page to have a footer etc., you should add it.
 
-{% code title="<plugin root>/src/Resources/config/services.xml" %}
-```markup
-<?xml version="1.0" ?>
+The `EventDispatcherInterface` is of course necessary in order to fire an event later on.
+
+Every page loader should implement a `load` method, which is not mandatory, but common sense. You want your page loader to work
```
mitelg, line 129:
```suggestion
Every page loader should implement a `load` method, which is not mandatory, but convention. You want your page loader to work
```

mitelg (changes requested):

guides/plugins/plugins/storefront/add-custom-page.md
PaddyS: And I'm not going to explain why I did that! :-P

PaddyS: Fixed

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+/**
+ * @RouteScope(scopes={"storefront"})
+ */
+class ExampleController extends StorefrontController
+{
+    /**
+    * @RouteScope(scopes={"storefront", "administration"})
```
mitelg, line 105: does this work? makes also no sense IMO 😁
Storefront route and admin scope?
maybe we should skip the entirely part about the route scope and we should really consider making an own article about that

mitelg (changes requested):

guides/plugins/plugins/storefront/add-custom-controller.md
PaddyS: I changed that because you didn't want the "custom-scope" to be in there.
Yet, I'd really like to have that part "multiple scopes possible" covered. So if you can come up with something better, go for it. :-P

mitelg: what about the idea to write an own article for the scope?
to be honest, I don't see a benefit to mention it here. I just want to implement a simple controller. custom or multiple route scopes seems more an advanced use case.

it should be mentioned that it is important to set this annotation because of this and that, but that should be sufficient for this guide, in my opinion

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+
+In order to add your own controller for your plugin, you first need a plugin as base.
+Therefore, you can refer to the [Plugin Base Guide](../plugin-base-guide.md).
+
+## Adding custom storefront controller
+
+### Storefront Controller class example
+
+First of all we have to create a new controller which extends from the `StorefrontController` class.
+A controller is also just a service which can be registered via the DI-Container.
```
lernhart, line 17:
```suggestion
A controller is also just a service which can be registered via the service container.
```
( Symfony documents their container as service container instead of DI-container )

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+    {
+        ...
+    }
+}
+```
+{% endcode %}
+
+### Services.xml example
+
+Next, we need to register our controller in the DI-container and make it public.
```
lernhart, line 118:
```suggestion
Next, we need to register our controller in the service container and make it public.
```

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+        https://symfony.com/schema/routing/routing-1.0.xsd">
+
+    <import resource="../../Storefront/Controller/**/*Controller.php" type="annotation" />
+</routes>
+```
+{% endcode %}
+
+### Adding template
+
+Now we registered our controller and Shopware indexes the route, but the template file, that is supposed to be rendered, is still missing.
```
lernhart, line 156:
```suggestion
Now that we registered our controller and Shopware indexes the route, but the template file, that is supposed to be rendered, is still missing.
```

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+Let's change that now.
+
+As previously mentioned, the code will try to render an `index.html.twig` file.
+Thus we have to create an `index.html.twig` in the `<plugin root>/src/Resources/views/storefront/page/example` directory, as defined in our controller.
+Below you can find an example, where we extend from the template `base.html.twig` and override the block `base_content`.
+In our [Customize templates guide](./customize-templates.md), you can learn more about customizing templates.
+
+{% code title="<plugin root>/src/Resources/views/storefront/page/example.html.twig" %}
+```text
```
lernhart, line 165: Why do some of you use \```text and some use \```twig ?
I think twig is working fine

```suggestion
```twig
```

guides/plugins/plugins/storefront/add-custom-controller.md
```diff
@@ -1,2 +1,216 @@
...
+    {
+        ...
+    }
+}
+```
+{% endcode %}
+
+## Next steps
+
+Now that you know how to create a custom controller, you can make it more beautiful with some styling.
```
lernhart, line 211:
```suggestion
Now that you know how to create a custom controller, you can make your templates shiny with some some styling.
```

lernhart (changes requested):

mitelg (approved):

guides/plugins/plugins/storefront/add-custom-controller.md
PaddyS: Afaik it only works fine because markdown defaults to text if unknown.
Twig is not a valid code markup known to default markdown, is it?

Sure, I can change that, won't break anything. Just... it will default to text anyway. What about HTML?

lernhart: @leichteckig had the same issue with using \```html, she changed it to \```twig. 
Dunno, if this matters at all.

PR #91, opened 2021-02-24
NEXT-12699  [NEW] Fill custom field with data

NiklasLimberg:
# What should be done?

Create a new article on our [GitBook instance|https://app.gitbook.com/@shopware/s/shopware/] which explains how to fill a custom field with data at runtime.

This article should mention:
- The prerequisite, a working plugin (Refer to the plugin base guide)
- Every other prerequisite you figure out during writing the guide (e.g. a subscriber, knowing how to create a service, a controller, etc.)
- Maybe the "Add custom field" guide as a prerequisite, so build upon it
- A short code example, including an explanation, on how to do it

Category: Extensions > Plugins > Framework > Custom field

## Are there already guides in the previous documentation for this?

Unfortunately, No. But please have a look at our example guide(e.g. note the PLACEHOLDERs for cross-references, that do not exist yet) and the Writing guidelines.

For more information, ask me, Patrick Stahl.


PaddyS (approved):

guides/plugins/plugins/using-custom-fields.md
```diff
@@ -0,0 +1,103 @@
...
+    </sw-custom-field-set-renderer>
+    <button @click="save">Save</button>
+</div>
+```
+
+## Loading the custom fields
+
+The following code sample shows how to load and save the the custom field. This code sample is tailored to the `sw-custom-field-set-renderer`, but it's not essential to use the `sw-custom-field-set-renderer`. The custom fields are loaded automatically and it's possible to just modify the `custom-fields` without loading the schema, because the schema is just needed for display purposes.
+
+```javascript
```
leichteckig, line 34: This example is quite big. Please use comments to emphasize the important lines. Or if are the two lines the only important ones, could we shorten the example a little bit? Only if possible :)

leichteckig, line 34: Please use Code Blocks

```
{% code title="path" %}
...
{% endcode %}
```

leichteckig (changes requested):

PaddyS:
Also, please rebase now :-)

guides/plugins/plugins/using-custom-fields.md
NiklasLimberg: i've added a few comments  i hope it's enough

PR #124, opened 2021-03-08
NEXT-13183 - fix code example

raknison:
added missing backticks


lernhart (approved):

guides/plugins/plugins/content/cms/add-data-to-cms-elements.md
```diff
@@ -0,0 +1,9 @@
+# Adding data to CMS elements
```
PaddyS, line 1: What happened here?

PaddyS (changes requested):

guides/plugins/plugins/content/cms/add-data-to-cms-elements.md
raknison: oops - accidentally commited that file

PR #135, opened 2021-03-10
NEXT-12583 - Add guide about changing line item prices in the cart

PaddyS:


guides/plugins/plugins/checkout/cart/change-price-of-item.md
```diff
@@ -0,0 +1,306 @@
...
+
+Now let's have a look inside the `process` method.
+
+We start by filtering all line items down to only products, just like we did in the `collect` method.
+Then we're iterating over all products found, building the unique key, which is necessary for fetching the new price from the `CartDataCollector`.
+
+If there's no price to be processed saved in the `CartDataCollector`, there's nothing to do here.
+Otherwise, we're fetching the new price, we're building a new instance of a `QuantityPriceDefinition` containing the new price.
+Using that instance, we can calculate the actual new price using the previously injected `QuantityPriceCalculator`.
+
```
lernhart, line 277: A hint could be useful here, stating that you should'nt do any external or database queries in the process method.
Even more, beacuse this is specifically asked in the trainings.

'Do not query the DB here. Use the collect method of the collector instead...'

lernhart (changes requested):

lernhart (approved):

PR #143, opened 2021-03-15
NTR - Remove links who cause format problems

leichteckig:


leichteckig:
Fixed by Gitbook themselves, so we don't need this workaround anymore. 🎉

PR #151, opened 2021-03-16
NEXT-12597 - Add guide about listening to order changes

PaddyS:


JoshuaBehrens:
Is this WIP? You have an example how to track order changes but no event on a state machine transition. I assumed that you'd listen to state changes using the state machine events.

PaddyS:
@JoshuaBehrens No, actually not WIP.
I still kinda don't feel like I need to an example for the order state events, since they're part of the class I already mentioned.

Any specific reason why I should add THAT event (or "those"), but not any other of the like 30 events in there?
Just struggling to see why exactly those are worth an extra mention? 🤔 

Totally up for discussion here!

PR #158, opened 2021-03-18
NEXT-12578 - Add guide about customizing the price calculation

PaddyS:


JoshuaBehrens:
Are there any docs about the quantity price calculator and the others? Maybe they should get a reference from here. I mean sure you can change the price but that calculation should be based on properly tested services like these

PR #164, opened 2021-03-23
NEXT-13857 - DAL prefix and suffix

Dominik28111:


shyim:
Wieso hat das ne eigene Seite?

PaddyS (changes requested):
Maybe should be part of `resources/references/core-reference/dal-reference/filters-reference.md` instead.

PR #193, opened 2021-04-01
NEXT-12722 - Using a modal window

raknison:
**What should be done?**
Create a new article on our GitBook instance which explains how to use a modal window in both types: Static modal and AJAX modal.
This article should mention:
The prerequisite, a working plugin (Refer to the plugin base guide)
Every other prerequisite you figure out during writing the guide (e.g. a subscriber, knowing how to create a service, a controller, etc.)
A short code example, including an explanation, on how to do it
An example each for static and ajax modal windows
Category: Extensions > Plugins > Storefront

### DIFF
--- a/guides/plugins/plugins/storefront/using-a-modal-window.md
+++ b/guides/plugins/plugins/storefront/using-a-modal-window.md
@@ -1,0 +1,368 @@
+# Using a modal window
+
+## Overview
+
+This guide explains how you can use a modal window in your plugin in different szenarios.
+
+## Prerequisites
+
+This guide requires you to already have a basic plugin running. This guide **does not** explain how to create a new plugin for Shopware 6. Head over to our Plugin base guide to learn how to create a plugin at first:
+
+{% page-ref page="./../plugin-base-guide.md" %}
+
+While this is not mandatory, having read the guide about adding custom JavaScript plugins beforehand might help you understand this guide a bit further:
+
+{% page-ref page="./add-custom-javascript.md" %}
+
+
+## Create a modal manually from the DOM using Boostrap
+
+The simples solution to create a modal is by using Boostrap. More info: [Modal Boostrap](https://getbootstrap.com/docs/4.0/components/modal/#live-demo)
+Here is a basic implementation as an example. We override the `base_main_inner` from the `@Storefront/storefront/page/content/index.html.twig` template to insert the modal specific DOM elements.
+
+{% code title="<plugin root>/src/Resources/views/storefront/page/content/index.html.twig" %}
+```text
+{% sw_extends '@Storefront/storefront/page/content/index.html.twig' %}
+
+{% block base_main_inner %}
+    <!-- Button trigger modal -->
+    <button type="button" class="btn btn-primary" data-toggle="modal" data-target="#exampleModal">
+        Launch demo modal
+    </button>
+
+    <!-- Modal -->
+    <div class="modal fade" id="exampleModal" tabindex="-1" role="dialog" aria-labelledby="exampleModalLabel" aria-hidden="true">
+        <div class="modal-dialog" role="document">
+            <div class="modal-content">
+                <div class="modal-header">
+                    <h5 class="modal-title" id="exampleModalLabel">Modal title</h5>
+                    <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+                        <span aria-hidden="true">&times;</span>
+                    </button>
+                </div>
+                <div class="modal-body">
+                    <!-- insert your content here -->
+                    ...
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
+                    <button type="button" class="btn btn-primary">Save changes</button>
+                </div>
+            </div>
+        </div>
+    </div>
+
+    {{ parent() }}
+{% endblock %}
+```
+{% endcode %}
+
+## Create a modal using AJAX Modal Utility
+
+When setting a `data-url` in addition to `data-toggle="modal"` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:
+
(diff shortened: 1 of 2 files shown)


guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,284 @@
+# Using a modal window
+
+## Overview
+
+This guide explains how you can use a modal window in your plugin.
```
tobiasberge, line 5: The guide itself looks good so far. 👍 

I am not sure if we already have another guide or explanation regarding a modal window, but here we only describe the "advanced" or programmatic way of doing this.

There are two more ways we could maybe describe here, and then in an "advanced" section we could describe the stuff you have already prepared.

### Create a modal manually from the DOM using Boostrap

The modal inside the shopware storefront is using the Bootstrap modal. More info: https://getbootstrap.com/docs/4.0/components/modal/#live-demo

```html
<!-- Button trigger modal -->
<button type="button" class="btn btn-primary" data-toggle="modal" data-target="#exampleModal">
  Launch demo modal
</button>

<!-- Modal -->
<div class="modal fade" id="exampleModal" tabindex="-1" role="dialog" aria-labelledby="exampleModalLabel" aria-hidden="true">
  <div class="modal-dialog" role="document">
    <div class="modal-content">
      <div class="modal-header">
        <h5 class="modal-title" id="exampleModalLabel">Modal title</h5>
        <button type="button" class="close" data-dismiss="modal" aria-label="Close">
          <span aria-hidden="true">&times;</span>
        </button>
      </div>
      <div class="modal-body">
        ...
      </div>
      <div class="modal-footer">
        <button type="button" class="btn btn-secondary" data-dismiss="modal">Close</button>
        <button type="button" class="btn btn-primary">Save changes</button>
      </div>
    </div>
  </div>
</div>
```

### Create a modal using Ajax Modal Utility

When setting a `data-url` in addition to `data-toggle="modal"` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:

```html
<button 
    class="btn"
    data-toggle="modal"
    data-url="https://example.org/ajax-url">
```

**HANDS UP: This does not work when the trigger selector is being changed via JavaScript, e.g. because of an ajax call which replaces the content**

### Advanced / manual using Pseudo Modal Utility

...

tobiasberge (changes requested):
Maybe we can include the other ways too

PaddyS (changes requested):

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,284 @@
...
+        // open the modal window and make it visible
+        this.modal.open();
+    }
+}
+```
+{% endcode %}
+
+## Closing the modal
+
+The `PseudoModalUtil` class also provide a `close()` method. As with opening the modal by calling `this.modal.open()`, you can simply close the modal with `this.modal.close()`.
```
leichteckig, line 121:
```suggestion
The `PseudoModalUtil` class also provide a `close()` method. Same as with opening the modal by calling `this.modal.open()`, you can simply close the modal with `this.modal.close()`.
```

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,284 @@
...
+
+## Callback when opening a modal
+
+The `open()` method of the `PseudoModalUtil` class supports a callback function as an argument. So, if you need to perform some action when your modal opens you can implement a callback like this:
```
leichteckig, line 125:
```suggestion
The `open()` method of the `PseudoModalUtil` class supports a callback function as an argument. So if you need to perform some action when your modal opens, you can implement a callback like this:
```

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,284 @@
...
+        this.modal = new PseudoModalUtil(content, useBackrop);
+        
+        // open the modal window and make it visible
+        this.modal.open();
+    }
+}
+```
+{% endcode %}
+
+As you can see in the sample code we are using the `js-pseudo-modal-template-title-element` class to style the title text of the modal.
```
leichteckig, line 250:
```suggestion
As you can see in the sample code, we are using the `js-pseudo-modal-template-title-element` class to style the title text of the modal.
```

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,284 @@
...
+It also tells the `PseudoModalUtil` class that the content of the `div` holds the title text.
+Furthermore there are 2 more css selectors `js-pseudo-modal-template` and `js-pseudo-modal-template-content-element` to define structure of the template string.
```
leichteckig, line 252:
```suggestion
Furthermore there are two more css selectors `js-pseudo-modal-template` and `js-pseudo-modal-template-content-element` to define the structure of the template string.
```

leichteckig (changes requested):

guides/plugins/plugins/storefront/using-a-modal-window.md
raknison: thanks @tobiasberge  - just added the missing parts

leichteckig (approved):

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -0,0 +1,368 @@
...
+            data-toggle="modal"
+            data-url="https://example.org/ajax-url">
+
+    {{ parent() }}
+{% endblock %}
+```
+{% endcode %}
+
+{% hint style="warning" %}
+This does not work when the trigger selector is being changed via JavaScript, e.g. because of an ajax call which replaces the content.
```
JoshuaBehrens, line 80: You write ajax here in small letters and later in capital. I think you want to write it in capitals.

JoshuaBehrens (changes requested):

PR #201, opened 2021-04-07
Add python3 samples

bpesch:


keulinho (commented):
Not a python 3 expert, but in general this looks good to me

The only problem i see is if we can maintain that script, as i think nobody of us is a python dev 

But i think it's totally valuable to have that kind of examples for different languages, maybe we should add a note above the code examples that only the PHP scrip is "officially supported" and guaranteed to work, the other languages are "community provided"

What do you think @PaddyS @elkmod ?

PaddyS:
Hey @bpesch,
thank you very much for your contribution! 🙏 

And yes, I do agree we need a note of some kind here.
Otherwise people might expect those kinds of code examples in every guide and kinda expect a consistency throughout all articles.

So if you could add that, that would be great:
```
{% hint style="warning" %}
Code examples written in languages other than PHP, Javascript or HTML are community-driven and not officially supported.
{% endhint %}
```

Maybe like that? @keulinho

bpesch:
> Otherwise people might expect those kinds of code examples in every guide and kinda expect a consistency throughout all articles.

If more contributions in other languages are expected, it may be suitable to setup tests to make sure each additional snippet has the same output as the PHP one, to make sure all snippets are working properly and consistent.

JoshuaBehrens:
I really like this addition from @bpesch . On the other hand this will get more debt into the documentation and this will easier getting out of date as shopware core devs are likely not to update it.

Maybe there can be some sort of community section which is still reviewed by shopware but more about community specific additions? This will make it easier to detect that content you read is not part of any shopware support stuff.

lernhart:
I agree with @JoshuaBehrens. Documentation of technologies out of range of our tech-stack can not be maintained properly. 

Maybe it makes sense to simply put a {% hint %} in the tab of the python example, rather than creating an own "community" section or similar, as @JoshuaBehrens suggested.

JoshuaBehrens:
@lernhart it feels like this will add up when more of these situations come up. So as long as some entity will be responsible for this it will not be marked as a hint (compare dockware docs). Maybe we can make friendsofshopware responsible for this on the long run.

sushmangupta:
@bpesch, Thanks for your contribution. :) 
The idea seems to be good, but the python samples once added for a concept, must be then added throughout the docs to maintain uniformity. Since the scope of these samples is just limited to a single app-base concept, we would like to dismiss this request.

PR #203, opened 2021-04-08
NEXT-14359 - Add the current sw version to all outgoing request to ex…

Tamvt:
We need to add the current sw version to all outgoing request to external apps.

### DIFF
--- a/guides/plugins/apps/administration/add-custom-module.md
+++ b/guides/plugins/apps/administration/add-custom-module.md
@@ -15,7 +15,7 @@ In order to create a custom module, you need to define an admin element to defin
         ...
     </meta>
     <admin>
-        <module name="exampleModule" source="https://example.com//promotion/view/promotion-module">
+        <module name="exampleModule" source="https://example.com/promotion/view/promotion-module">
             <label>Example Module</label>
             <label lang="de-DE">Beispiel Modul</label>
         </module>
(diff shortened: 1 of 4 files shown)


keulinho:
Please also update
https://github.com/shopware/docs/blob/master/guides/plugins/apps/app-base-guide.md#registration-request
https://github.com/shopware/docs/blob/master/guides/plugins/apps/app-base-guide.md#confirmation-request
https://github.com/shopware/docs/blob/master/guides/plugins/apps/app-base-guide.md#webhooks
https://github.com/shopware/docs/blob/master/guides/plugins/apps/app-base-guide.md#api-docs
https://github.com/shopware/docs/blob/master/guides/plugins/apps/administration/add-custom-action-button.md

And mention that the current shopware version is sent as `sw-version` header.

The api-doc section may be a bit tricky, as it is some custom gitbook format for documenting apis, but you should be able to use the existing doc for the signature header and copy and paste that:
```
{% api-method-parameter name="shopware-app-signature" type="string" required=true %} The hmac-signature of the query string, signed with the app secret {% endapi-method-parameter %}
```

guides/plugins/apps/administration/add-custom-module.md
```diff
@@ -32,12 +32,13 @@ If the user opens the module in the administration your app will receive a reque
...
 * shop-url: The URL of the shop, this can later be used to access the Shopware API
 * timestamp: The Unix timestamp when the request was created
+* sw-version: the current version of the shopware that the app installed on.
 * shopware-shop-signature: sha256 hmac of the rest of the query string, signed with the `shop-secret`
 
 A sample request may look like this:
 
 ```text
-https://example.com//promotion/view/promotion-config?shop-id=HKTOOpH9nUQ2&shop-url=http%3A%2F%2Fmy.shop.com&timestamp=1592406102&shopware-shop-signature=3621fffa80187f6d43ce6cb25760340ab9ba2ea2f601e6a78a002e601579f415
+https://example.com//promotion/view/promotion-config?shop-id=HKTOOpH9nUQ2&shop-url=http%3A%2F%2Fmy.shop.com&timestamp=1592406102&sw-version=6.4.9999999.9999999-dev&shopware-shop-signature=3621fffa80187f6d43ce6cb25760340ab9ba2ea2f601e6a78a002e601579f415
```
JoshuaBehrens, line 41: I suggest to remove the double slash after the domain of this url

JoshuaBehrens (approved):
Just a suggestion as this is not the same topic as this PR but at the same line

lernhart (approved):

keulinho:
the changes were made to current trunk, meaning they will be released with 6.4.1.0

This was merged to master, which is 6.3 as of now right?

We should take of this in the future to at least merge it to the right major version branch.

And how do we handle changes/feature that are made in minor version? Like the feature documented here? Do we need to take that into account in the documentation text itself?

keulinho:
Sorry i didn't thought about the versioning stuff before for this PR.

leichteckig:
Hello, that's alright and rightfully so - no need to excuse. 

We do have versioning possibilities in the documentation but it's not fully implemented yet. If it's a 6.4.0.0 only topic, the branch `6.4.0.0` needs to be used. I'm not sure how often it's rebased, though.

In this case, I think we need to revert / undo this change. I'll take care of that. Thank you for this notification 👍

keulinho:
It's the documentation of a feature introduced in 6.4.1.0.

If we have branches for all minor updates we should target the 6.4.1.0 branch (if it already exists).

Otherwise we could add a hint in every sentence like:
```
Since shopware 6.4.1.0 the current shopware version will be sent as a `sw-version` header.
```

I'm ok with both approaches, but i guess we will run into things like this regularly (especially for the app-system where we are actively adding features that are always relevant for the developers).

Maybe we clarify here how to handle such cases in general.

PR #213, opened 2021-04-12
NTR - Improve guide on adding tabs

leichteckig:
With this MR I want to improve the tab guide a bit, by tackling the following issues:

* Highlight the paths
*  Improve the structure
* Fixing minor language issues
* Use Gitbook syntax

If you spot more possible improvements, please don't hesitate to mention them here. 💙

### DIFF
--- a/guides/plugins/plugins/administration/add-new-tab.md
+++ b/guides/plugins/plugins/administration/add-new-tab.md
@@ -72,7 +72,8 @@ Unfortunately, you cannot use the block mentioned above, because then your new t
 
 Knowing the block you have to override in your plugin, you can now start doing exactly this: Add your custom tab by overriding this block called `sw_product_detail_content_tabs_reviews`.
 
-{% hint style="danger" %} However, please keep in mind that "overriding" doesn't mean we want to replace the block completely with our new one. We want to add our tab, thus only extending the template. This will have some implications on our implementation.
+{% hint style="danger" %}
+However, please keep in mind that "overriding" doesn't mean we want to replace the block completely with our new one. We want to add our tab, thus only extending the template. This will have some implications on our implementation.
 {% endhint %}
 
 First, please re-create the directory structure from the core code in your plugin. In this case, you'll have to create a directory structure like the following: `<plugin root>/src/Resources/app/administration/src/page/sw-product-detail`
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/administration/add-new-tab.md
```diff
@@ -120,14 +165,17 @@ Shopware.Module.register('sw-new-tab-custom', {
     }
 });
 ```
+{% endcode %}
 
 As already mentioned, you need to create a dummy module in order to override the `routeMiddleware` method. In there, you're listening for the current route, that got called. If the current route matches `sw.product.detail`, you want to add your new child route to it, and that's what's done here.
 
-Your child route defines the routes name, so make sure to use the name you're already defined earlier! The path should be identical to the default ones, which look like this: `/sw/product/detail/:id/base` Just replace the `base` here with `custom` or anything you like.
+{% hint style="warning" %} Your child route defines the routes name, so make sure to use the name you're already defined earlier!  {% endhint %}
```
NiklasLimberg, line 172:
```suggestion
{% hint style="warning" %} 
Your child route defines the routes name, so make sure to use the name you're already defined earlier!
{% endhint %}
```

guides/plugins/plugins/administration/add-new-tab.md
```diff
@@ -47,40 +64,66 @@ Having a look at the template, you might find the block `sw_product_detail_conte
...
+
+    {# This parent is very important as you don't want to override the review tab completely #}
+    {% parent %}
+
+{% endblock %}
+```
+{% endcode %}
+
+{% hint style="warning" %} The block gets overridden and immediately the parent block is called, since you do not want to replace the 'Review' tab, you want to add a new tab instead.
+{% endhint %}
```
NiklasLimberg, line 108:
```suggestion
{% hint style="warning" %}
The block gets overridden and immediately the parent block is called, since you do not want to replace the 'Review' tab, you want to add a new tab instead.
{% endhint %}
```

guides/plugins/plugins/administration/add-new-tab.md
```diff
@@ -47,40 +64,66 @@ Having a look at the template, you might find the block `sw_product_detail_conte
...
+
+### Create custom tab
 
-Unfortunately, you cannot use the block mentioned above, because then your new tab wouldn't be inside the `<sw-tabs>` element. Instead, you can choose the last available block inside the element, which is `sw_product_detail_content_tabs_reviews`.
+Knowing the block you have to override in your plugin, you can now start doing exactly this: Add your custom tab by overriding this block called `sw_product_detail_content_tabs_reviews`.
 
-Knowing the block you have to override in your plugin, you can now start doing exactly this. First re-create the directory structure from the core code in your plugin. In this case, you'll have to create a directory structure like the following: `<plugin root>/src/Resources/app/administration/src/page/sw-product-detail`
+{% hint style="danger" %} However, please keep in mind that "overriding" doesn't mean we want to replace the block completely with our new one. We want to add our tab, thus only extending the template. This will have some implications on our implementation.
+{% endhint %}
```
NiklasLimberg, line 76:
```suggestion
{% hint style="danger" %}
However, please keep in mind that "overriding" doesn't mean we want to replace the block completely with our new one. We want to add our tab, thus only extending the template. This will have some implications on our implementation.
{% endhint %}
```

NiklasLimberg (changes requested):

lernhart (approved):
Awesome !

NiklasLimberg (approved):

PR #243, opened 2021-04-24
Remove example statement

JoshuaBehrens:
It is not an example. It is the way you do it. Documents belong to orders regarding DDL

```
`order_id` BINARY(16) NOT NULL,
`order_version_id` BINARY(16) NOT NULL,
```


PaddyS:
Thank you @JoshuaBehrens 🙏🏻

JoshuaBehrens:
Thank you for merging :)

PR #248, opened 2021-04-27
NEXT-14805 - Update docs for jest helper

jleifeld:
The jest guide does not contain the new helper methods for the administration tests. I added these so it now matches the README file again.


leichteckig (changes requested):

guides/plugins/plugins/testing/jest-admin.md
```diff
@@ -347,18 +347,115 @@ Vue Test Utils has some advanced features for stubbing components. A stub is act
...
+    expect(wrapper.vm).toBeTruthy();
+});
+```
+
+### Repository factory
+The data handling and the repository factory works by default. It will be generated by the entity-schema which will be written to a file before you start the test suite.
+
+Everytime the repository factory request something from a URL you get a notification in the console. This notification also includes a short guide how to implement the response. This information could look something like this:
+
+```
```
lernhart, line 389:
``````suggestion
```js
``````

lernhart (changes requested):

leichteckig (approved):

lernhart (approved):

PR #255, opened 2021-05-06
NEXT-14679 - Webhook Preventing replay attacks

nguyenngon95:


guides/plugins/apps/app-base-guide.md
```diff
@@ -302,6 +303,8 @@ Where the `source` property contains all necessary information about the Shopwar
 
 The next property `data` contains the name of the event so that a single endpoint can handle several different events, should you desire. `data` also contains the event data in the `payload` property, due to the asynchronous nature of theses webhooks the `payload` for `entity.written` events does not contain complete entities as these might become outdated. Instead the entity in the payload is characterized by its id, stored under `primaryKey`, so that the app can fetch additional data through the shops API. This also has the advantage of giving the app explicit control over the associations that get fetched instead of relying on the associations determined by the event. Other events in contrast contain the entity data that defines the event, but keep in mind that event might not contain all associations.
 
+The next property `timestamp` is the time which the webhook was handled. The attacker cannot change the timestamp without validating the signature. If the timestamp is too old, the user's application can decide to reject the request.
```
keulinho, line 306: What do you think of:
```suggestion
The next property `timestamp` is the time which the webhook was handled. This can be used to prevent replay attacks, as an attacker cannot change the timestamp without making the signature invalid. If the timestamp is too old, your app should reject the request.
```

keulinho (changes requested):

guides/plugins/apps/app-base-guide.md
keulinho: Additionally pleas make clear that this property is only available from 6.4.1.0 onwards

keulinho (approved):

PR #260, opened 2021-05-06
Update add-custom-component.md

be-codified:
Please see comment in code.

### DIFF
--- a/guides/plugins/plugins/administration/add-custom-component.md
+++ b/guides/plugins/plugins/administration/add-custom-component.md
@@ -128,21 +128,30 @@ As mentioned above, Shopware 6 looks for a `main.js` file in your plugin. Its co
 Your plugin has to be activated for this to work.
 {% endhint %}
 
-Make sure to also include that file when publishing your plugin! A copy of this file will then be put into the directory `<shopware root>/public/bundles/customcomponent/administration/js/custom-component.js`.
+<!--
+    NOTES to bottom part of code: 
+    - I have tried to include file in Shopware v6.3.5.2 but including or having this twig file has no effect. It works out of the box. Can you please recheck this part of documentation?
+    - There is strange bug when using name e.g. CustomBundle as it will ignore word Bundle and compile folder name just `custom` instead of `custombundle`, then SW will not find it to load. Though it works with name e.g. CustomBundles (note the s letter).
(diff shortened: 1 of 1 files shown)


PaddyS:
Hey @be-codified,

thank you for your contribution! 🎉 
However, can you please remove your comment and resolve this issue by removing the whole "## Loading the JS files" section? :-)

You're absolutely right, this is not necessary anymore.

PaddyS:
Hey @be-codified,

so I had a look at the said issue about the "CustomBundle" bug.
I created a new plugin called "CustomBundle" and created a simple administration file.

I then build the administration and even tried the watcher. In both cases you are right about the "bundle" disappearing from the folder-name, but the file is still being found and executed for me.

Can you confirm that's the case? Maybe that was only an issue when you were trying to manually load the public file?
You know, like in the section you just removed. 🤔

be-codified:
@PaddyS it is kinda blurry to me as we working on this some time ago.

But I have the feeling that the file wasn't found. This is why we couldn't proceed and then we figured out about missing bundle string in folder name.

How does the system checks for file(s)? Does it go through the folder()s or it has folder(s) path written in db? In our case it was seeking for CustomBundle instead of Custom.

I hope this helps you.

PaddyS:
Hey @be-codified,

I just double-checked and definitely can't find any real issues here, besides the `bundle` being removed from the directory.
[Here](https://github.com/shopware/platform/blob/v6.4.0.0/src/Core/Framework/Api/Controller/InfoController.php#L196) you can see that only "bundle" as a word is being dropped on intention.

The plugins are being loaded [here](https://github.com/shopware/platform/blob/v6.4.0.0/src/Administration/Resources/app/administration/src/core/application.js#L510-L524).
The said [sw-plugin-dev.json file](https://github.com/shopware/platform/blob/v6.4.0.0/src/Administration/Resources/app/administration/webpack.config.js#L523-L540) is created by the [webpack assets plugin](https://www.npmjs.com/package/assets-webpack-plugin).

be-codified:
@PaddyS I trust you :)

PaddyS:
@be-codified Aw, thanks. 🙏🏻

If you're still running into that issue one day, I'd be glad to see a new issue here on GitHub, so we can have a deeper look on any possible side-effects. :-)

PR #265, opened 2021-05-10
Video embeddings for documentation

elkmod:


guides/installation/dockware.md
```diff
@@ -150,3 +150,9 @@ Would you like to explore alternative ways to install Shopware? You can install
 
 Otherwise, you might want to start writing your very own plugin. Head over to [Plugin base guide](../plugins/plugins/plugin-base-guide.md) to get a grip on that topic.
 
+
+{% hint style="info" %}
+Here's a video explaining the basics of Dockware from our free online training ["Backend Development"](https://academy.shopware.com/courses/shopware-6-backend-development-with-jisse-reitsma).
+
+**[Using Dockware](https://www.youtube.com/watch?v=b9wVfUOKqmI)**
```
leichteckig, line 157: When it comes to youtube videos in general, isn't it possible to use `embed` Gitbook syntax here? 🙂

PaddyS: He's doing that later on. Seems like it's on intention? 🤔

leichteckig: Possibly - Abd I'm aware of that, but I think it would be more comfortable for the reader to use those elements whenever possible. 🙂

elkmod: The idea was to use either embed or inline references based on which one fits better. In some cases (like in that one), I felt an embed would steal too much attention :)

leichteckig: Fair enough 👍

PR #278, opened 2021-05-14
NEXT-15001 - Add backward compatibility guide

tobiasberge:
Add backward compatibility guide to:

```
- guidelines
    - code
        -  backward-compatibility
```

- Replaced internal GitLab links with GitHub documentation and ADRs
- Replaced GitLab versioning strategy with shopware.com blog post
- Replaced GitLab emoji codes with UTF-8 emojis

@taltholtmann @elkmod Do you have experience with plain HTML tables in GitBook? Does this work when there are e.g. Markdown links inside the table? Should we replace it with markdown tables?

### DIFF
--- a/resources/guidelines/code/backward-compatibility.md
+++ b/resources/guidelines/code/backward-compatibility.md
@@ -139,14 +139,14 @@ As Shopware is based on the PHP framework Symfony, we also have to make sure to 
 
 #### JavaScript
 
-| Use Case                                           | Allowed? | Notes / Alternatives                                                                               |
-| ---                                                | ---      | ---                                                                                                |
-| Renaming or removing JS services.                  | 🔴 NO    |                                                                                                    |
-| Renaming or removing of JS plugins.                | 🔴 NO    |                                                                                                    |
-| Changing the public API of a JS plugin or service. | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Add new public function](#add-new-public-function) |
-| Renaming methods of JS plugins or services.        | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Rename a method](#rename-a-method)                 |
-| Renaming or removing of JS events.                 | 🔴 NO    |                                                                                                    |
-| Changing the parameters of JS events.              | 🔴 NO    |                                                                                                    |
+| Use Case                                           | Allowed? | Notes / Alternatives                                                                                                  |
+| ---                                                | ---      | ---                                                                                                                   |
+| Renaming or removing JS services.                  | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Renaming or removing JS services](#renaming-or-removing-js-services)  |
+| Renaming or removing of JS plugins.                | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Renaming or removing JS plugins](#renaming-or-removing-js-plugins) |
+| Changing the public API of a JS plugin or service. | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Add new public function](#add-new-public-function)                    |
+| Renaming methods of JS plugins or services.        | 🔴 NO    | Use the deprecation workflow.<br>Code Example: [Rename a method](#rename-a-method)                                    |
+| Renaming or removing of JS events.                 | 🔴 NO    |                                                                                                                       |
+| Changing the parameters of JS events.              | 🔴 NO    |                                                                                                                       |
 
 #### Styling / CSS
 
@@ -403,5 +403,39 @@ Shopware.Component.register('sw-old', {
             );
         }
     }
+}
+```
+
+#### Renaming or removing JS services
+
+```javascript
+// http-client.service.js
+/**
+* @deprecated tag:v6.5.0 - Use NewHttpClient instead (new-http-client.service.js)
+*/
+export default class HttpClient {
+    // ...
+}
+ 
+// new-http-client.service.js
+export default class NewHttpClient {
+    // ...
+}
+```
+
+#### Renaming or removing JS plugins
+
+```javascript
+// buy-button.plugin.js
+/**
+* @deprecated tag:v6.5.0 - Use NewBuyButtonPlugin instead (new-buy-button.plugin.js)
+*/
+export default class BuyButtonPlugin extends Plugin {
+    // ...
+}
+ 
+// new-buy-button.plugin.js
+export default class NewBuyButtonPlugin extends Plugin {
+    // ...
 }
 ```
--- a/resources/guidelines/code/backward-compatibility.md
+++ b/resources/guidelines/code/backward-compatibility.md
@@ -1,590 +1,213 @@
 # Backward Compatibility
 
 ## Introduction
-Shopware is a standard eCommerce solution which is used by many customers around the world and is the technical foundation of their online business. As developers of Shopware it is our highest goal to guarantee the reliability of the software for our customers. Always keep in mind that every change you make can have a big impact on the live of our customers, either in a good way, but also in a bad way.
+
+Shopware is a standard eCommerce solution which is used by many customers around the world and is the technical foundation of their online business. As developers of Shopware it is our highest goal to guarantee the reliability of the software for our customers. Always keep in mind that every change you make can have a big impact on the life of our customers, either in a good way, but also in a bad way.
 
 For the release strategy Shopware uses a semantic [versioning](https://www.shopware.com/en/news/shopware-6-versioning-strategy/) and therefore has to be always backward compatible for minor and patch updates. This brings additional challenges, when changing the code of Shopware. The following guide aims to provide you with the necessary workflows and techniques to do your changes in a backward compatible way and what to do, if it is not possible.
 
 ## Annotations
+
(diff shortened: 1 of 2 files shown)


elkmod:
> Add backward compatibility guide to:
> 
> ```
> - guidelines
>     - code
>         -  backward-compatibility
> ```
> 
> * Replaced internal GitLab links with GitHub documentation and ADRs
> * Replaced GitLab versioning strategy with shopware.com blog post
> * Replaced GitLab emoji codes with UTF-8 emojis
> 
> @taltholtmann @elkmod Do you have experience with plain HTML tables in GitBook? Does this work when there are e.g. Markdown links inside the table? Should we replace it with markdown tables?

@tobiasberge Markdown tables will work better, so rather convert them. I will add some more comments, mostly regarding formatting.

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
+# Backward Compatibility
+
+## Introduction
+Shopware is a standard eCommerce solution which is used by many customers around the world and is the technical foundation of their online business. As developers of Shopware it is our highest goal to guarantee the reliability of the software for our customers. Always keep in mind that every change you make can have a big impact on the live of our customers, either in a good way, but also in a bad way.
```
elkmod, line 4: Blank space before paragraphs is mandatory otherwise, Gitbook will have display issues.

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+
+For the release strategy Shopware uses a semantic [versioning](https://www.shopware.com/en/news/shopware-6-versioning-strategy/) and therefore has to be always backward compatible for minor and patch updates. This brings additional challenges, when changing the code of Shopware. The following guide aims to provide you with the necessary workflows and techniques to do your changes in a backward compatible way and what to do, if it is not possible.
+
+## Annotations
+During the development different cases occur, where you want to replace old code with new one or even remove some obsolete code. As Shopware must always be backward compatible on minor and patch updates, old code should only be removed with a major release. Until the next major release you want to mark the code with a corresponding annotation, to inform other developers of the planned change. This overview shows the most important annotations and in which situation they have to be used.
```
elkmod, line 9: See above.

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+
+### @deprecated
+```php
```
elkmod, line 12: For every heading :)

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+/**
+ * @deprecated tag:v6.5.0 - Use NewFunction() instead
+ */
+```
```
elkmod, line 16: Also here one more whitespace

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+ * @internal (flag:FEATURE_NEXT_11111)
+ */
+```
+The `@internal` annotation is used for new introduced code, which is not yet released or hidden behind a feature flag. This ensures, that it will not be treated as public API until the corresponding feature is released and makes it possible to change the code in any way until the final release. Always add the name of the corresponding feature flag to the annotation, so that it will not be forgotten, when the corresponding feature is released.
+
+## Workflows
+### Backward Compatible Features
+When developing new features, the goal should always be, to do this in a backward compatible way. This ensures that the feature can be shipped with a minor release, to provide value for customers as soon as possible. The following table should help you to use the correct approach for each type of change.
+
+<table>
```
elkmod, line 47: Markdown-Table. [Gitbook Markdown](https://docs.gitbook.com/editing-content/markdown).

elkmod (commented):
Is it reasonable to split this guide into multiple files/articles? There are 4 levels of nested headings, so maybe have the second-level ones as separate ones?

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+
+#### Deprecate CSS selectors
+```HTML
+{# @deprecated tag:v6.5.0 - CSS class "card-primary" is deprecated, use "card-major" instead #}
+<div class="card card-major card-primary">
+    ...
+</div>
+```
+
+## JavaScript
```
tobiasberge, line 685: The heading level is not correct here, I will fix it

tobiasberge:
> Is it reasonable to split this guide into multiple files/articles? There are 4 levels of nested headings, so maybe have the second-level ones as separate ones?

@elkmod Sure we could do that. It is mostly copy pasta from GitLab right now where it is all on one page with an extra table of contents. In my mind this was okay because of the generated sidebar navigation in GitBook.

But how would we deal with the introduction? Should we do it as a separate article or in the README.md? 

```
- backward-compatibility
    - README.md (Introduction)
    - Annotations
    - Workflows
    - Compatibility Sheet
    - Code examples
```
But I'm not sure here about the code examples which are linked, maybe this would be harder to follow when the code examples are on a different page.

resources/guidelines/code/backward-compatibility.md
PaddyS, line 4:
```suggestion
Shopware is a standard eCommerce solution which is used by many customers around the world and is the technical foundation of their online business. As developers of Shopware it is our highest goal to guarantee the reliability of the software for our customers. Always keep in mind that every change you make can have a big impact on the life of our customers, either in a good way, but also in a bad way.
```
```suggestion
Shopware is a standard eCommerce solution which is used by many customers around the world and is the technical foundation of their online business. As developers of Shopware it is our highest goal to guarantee the reliability of the software for our customers. Always keep in mind that every change you make can have a big impact on the live of our customers, either in a good way, but also in a bad way.
```

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+        <td>⚪ PARTIAL</td>
+        <td>Only optional arguments are allowed to be added and this should be made via <code>func_get_args()</code>.<br><br>Code Example: [Add an argument](#add-an-argument)</td>
+    </tr>
+    <tr>
+        <td>Change the arguments of a private method.</td>
+        <td>✅ YES</td>
+        <td></td>
+    </tr>
+    <tr>
+        <td>Change return the type of a method.</td>
```
PaddyS, line 170:
```suggestion
        <td>Change the return type of a method.</td>
```
I suppose?

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+<table>
+	<tr>
+		<th width=260>Use Case</th>
+		<th width=120>Allowed?</th>
+		<th>Notes / Alternatives</th>
+	</tr>
+	<tr>
+		<td>Renaming or removing JS services.</td>
+		<td>🔴 NO</td>
+		<td></td>
```
PaddyS, line 340: Doesn't the deprecation workflow apply here as well? And the same for "Renaming or removing of JS plugins"?

resources/guidelines/code/backward-compatibility.md
```diff
@@ -0,0 +1,751 @@
...
+	</tr>
+	<tr>
+		<td>Renaming methods of JS plugins or services.</td>
+		<td>🔴 NO</td>
+		<td>Use the deprecation workflow.<br><br>Code Example: [Rename a method](#rename-a-method)</td>
+	</tr>
+	<tr>
+		<td>Renaming or removing of JS events.</td>
+		<td>🔴 NO</td>
+		<td></td>
```
PaddyS, line 360: Same here: Deprecation workflow wouldn't work here?

PaddyS (changes requested):

PaddyS:
@elkmod @tobiasberge I do think leaving it as one file is better.

All the information on one page, instead of having to browse through multiple pages. Don't think the headlines will be an issue. 🤔

resources/guidelines/code/backward-compatibility.md
tobiasberge: Yes there was an example in the past, not sure why this wasn't in the GitLab version. Will try to clear this up.

resources/guidelines/code/backward-compatibility.md
tobiasberge: See: https://github.com/shopware/docs/pull/278#discussion_r634978337

resources/guidelines/code/backward-compatibility.md
tobiasberge: I've converted all tables to markdown, it just looks broken in the github diff because of the enforced soft-wrap of long lines.

resources/guidelines/code/backward-compatibility.md
tobiasberge: I added a code example for this case and for other cases. For those ones which are missing we cannot really recommend an "easy/straight-forward" deprecation way because it too highly depends on the specific case (e.g. the route renaming, or renaming entire files etc.)

"No" in this document also doesn't mean that it's impossible/forbidden or something. It just means, that the developer cannot "just change" these things or simply put "deprecated" on it. It can be possible to make it backward compatible even when the list says "no". It's just not really possible to define a generic example which applies to all possible cases. Not sure how we can make that more clear. It should help the developer to find out what can be done easily and what needs closer attention. 

For example the "data props" in the Vue components. Those can contain dumb booleans or entire entity collections which get loaded via the API. Yes it is possible to deprecate those data props but they could be used in the template, in methods, in computed properties and so on and it would not be enough to say "Add deprecation to the property inside `data`" And the code example would almost be a small module.

PR #280, opened 2021-05-18
NEXT-14408 - Fix page references

fschmtt:
I noticed faulty page references that are fixed with this PR.

Also, they weren't identified by the `Check Markdown links` workflow.


PaddyS:
Will merge if it's green. :-)

PR #281, opened 2021-05-18
Change ProductCollector to ProductCartProcessor

juanitolaguna:
There is no such class ProductColllector. But the ProductCartProcessor is implementing the CartDataCollectorInterface. Changed that too btw.


PaddyS:
Thank you very much @juanitolaguna 🎉

PR #288, opened 2021-05-21
Translations: Added a section on general use in PHP

Josniii:
I noticed there was a section on using translation inside PHP controllers, but nothing noted about general use, so I figured I'd add it, so it's more straight forward if your use-case isn't translating directly in a controller. I spent a bit of time looking for how to do this the first time I needed to do translations in PHP, so I hope this can help new developers @ WEXO and elsewhere save some time getting up to speed on translation usage.

If you think it needs changes, feel free to let me know.

### DIFF
--- a/guides/plugins/plugins/storefront/add-translations.md
+++ b/guides/plugins/plugins/storefront/add-translations.md
@@ -82,11 +82,11 @@ $this->trans('soldProducts', ['%count%' => 3, '%country%' => 'Germany']);
 
 ## Using translation generally in PHP
 
-If we need to use a snippet elsewhere in PHP, we can use Dependency Injection to inject `Shopware\Core\Framework\Adapter\Translation\Translator` which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
+If we need to use a snippet elsewhere in PHP, we can use Dependency Injection to inject `translator` which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
 
 ```xml
 <service id="Swag\Example\Service\SwagService" public="true" >
-    <argument type="service" id="Shopware\Core\Framework\Adapter\Translation\Translator" />
+    <argument type="service" id="translator" />
 </service>
 ```
 
--- a/guides/plugins/plugins/storefront/add-translations.md
+++ b/guides/plugins/plugins/storefront/add-translations.md
@@ -82,7 +82,7 @@ $this->trans('soldProducts', ['%count%' => 3, '%country%' => 'Germany']);
 
 ## Using translation generally in PHP
 
-If we need to use a snippet elsewhere in PHP, we can use Dependency Injection to inject `translator` which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
+If we need to use a snippet elsewhere in PHP, we can use [Dependency Injection](../plugin-fundamentals/dependency-injection.md) to inject the `translator`, which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
 
 ```xml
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/storefront/add-translations.md
```diff
@@ -80,3 +80,27 @@ Translation with placeholders:
 $this->trans('soldProducts', ['%count%' => 3, '%country%' => 'Germany']);
 ```
 
+## Using translation generally in PHP
+
+If we need to use a snippet elsewhere in PHP, we can use Dependency Injection to inject `Shopware\Core\Framework\Adapter\Translation\Translator` which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
+
+```xml
+<service id="Swag\Example\Service\SwagService" public="true" >
+    <argument type="service" id="Shopware\Core\Framework\Adapter\Translation\Translator" />
```
JoshuaBehrens, line 89: I think you should refer to the `translator` service id instead as the adapter here is just a decorator for that service.

JoshuaBehrens (changes requested):

guides/plugins/plugins/storefront/add-translations.md
Josniii: Done. I hadn't even noticed that myself, I've even seen several store plugins that use the Adapter directly.

JoshuaBehrens: 👀 I don't think that this is a good idea but hey good that the docs are better

JoshuaBehrens (approved):

guides/plugins/plugins/storefront/add-translations.md
```diff
@@ -80,3 +80,27 @@ Translation with placeholders:
 $this->trans('soldProducts', ['%count%' => 3, '%country%' => 'Germany']);
 ```
 
+## Using translation generally in PHP
+
+If we need to use a snippet elsewhere in PHP, we can use Dependency Injection to inject `translator` which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
```
PaddyS, line 85:
```suggestion
If we need to use a snippet elsewhere in PHP, we can use [Dependency Injection](../plugin-fundamentals/dependency-injection.md) to inject the `translator`, which implements Symfony's `Symfony\Contracts\Translation\TranslatorInterface`:
```

PaddyS (changes requested):

guides/plugins/plugins/storefront/add-translations.md
Josniii: Me neither. But when the core itself uses the Adapter (https://github.com/shopware/core/blob/trunk/Checkout/DependencyInjection/document.xml#L71), that's what you find in the code and end up copying when the docs don't mention it 😄

guides/plugins/plugins/storefront/add-translations.md
Josniii: Added, thanks!

PaddyS (approved):

guides/plugins/plugins/storefront/add-translations.md
JoshuaBehrens: Will you or do I PR that away?

Josniii: I found it quite a few places, I'll make a PR that removes it 😄

JoshuaBehrens: Link it to me :D

Josniii: PR is here:
https://github.com/shopware/platform/pull/1884

PaddyS:
Thank you very much @Josniii 🎉

PR #291, opened 2021-05-25
NEXT-12673 - Add data indexer guide

taltholtmann:


guides/plugins/plugins/framework/data-handling/add-data-indexer.md
```diff
@@ -0,0 +1,263 @@
+# Adding data indexer
+
+## Overview
+
+Data indexer are used to optimize the performance of recurring complex tasks. One good example to understand the benefit of data indexer would be the cheapest price calculation within Shopware. Every product has a `cheapest_price` column in the database which should contain the cheapest price a product has. The calculation of this column can be complex, because a product can have several variants with advanced pricing rules and so on. This makes the calculation more difficult and would take too much time when reading 25 products for a listing. To optimize the performance there is a data indexer that calculates the cheapest price of a product every time the product gets any changes in the database. This means that no new calculation has to be performed when a product is read, and performance during reading is significantly increased. Furthermore data indexer can make use of the [message queue](../../../../hosting/infrastructure/message-queue.md) to handle the calculations asynchronously.
```
pweyck, line 5:
```suggestion
Data indexer are used to optimize the performance of recurring complex tasks. One good example to understand the benefit of data indexer would be the cheapest price calculation within Shopware. Every product has a `cheapest_price` column in the database which should contain the cheapest price a product has. The calculation of this column can be complex, because a product can have several variants with advanced pricing rules and so on. This makes the calculation more difficult and would take too much time when reading 25 products for a listing. To optimize the performance there is a data indexer that calculates the cheapest price of a product every time the product is updated by the DAL. This means that no new calculation has to be performed when a product is read, and performance during reading is significantly increased. Furthermore data indexer can make use of the [message queue](../../../../hosting/infrastructure/message-queue.md) to handle the calculations asynchronously.
```

guides/plugins/plugins/framework/data-handling/add-data-indexer.md
```diff
@@ -0,0 +1,263 @@
...
+
+## Prerequisites
+
+This guide is build upon the [plugin base guide](../../plugin-base-guide.md), but any plugin will work here. Just note that all examples are using the plugin mentioned above. In order to create data indexer you should have read the [adding complex data guide](./add-custom-complex-data.md).
+
+## Adding an own data indexer
+
+It is possible to add data indexer for own entities, like the one created in the [adding complex data guide](./add-custom-complex-data.md) or adding indexer for existing entities. However, if you want to react on changes of existing entities the preferred way should be subscribing to the events if available. See the [Index data using existing events](Index data using existing events) section below. To create a new indexer, just create a new class in your plugin:
```
pweyck, line 13:
```suggestion
It is possible to add data indexer for your own entities, like the one created in the [adding complex data guide](./add-custom-complex-data.md) or for existing entities. However, if you want to react on changes of existing entities the preferred way should be subscribing to the events if available. See the [Index data using existing events](Index data using existing events) section below. To create a new indexer, just create a new class in your plugin:
```

guides/plugins/plugins/framework/data-handling/add-data-indexer.md
```diff
@@ -0,0 +1,263 @@
...
+    <services>
+        <service id="Swag\BasicExample\Service\Subscriber" >
+            <argument type="service" id="Doctrine\DBAL\Connection" />
+            <tag name="kernel.event_subscriber" />
+        </service>
+    </services>
+</container>
+```
+
+It is recommended to work directly with the `Connection` since the event is dispatched in the context of an indexer. If we would use the Data Abstraction Layer \(DAL\) for writing changes to the database, the indexer could be triggered again, because it listens for `EntityWrittenContainerEvent` events. This could lead to an infinite loop. Using the `Connection` directly prevents the DAL from dispatching entity written events.
```
pweyck, line 262:
```suggestion
It is recommended to work directly with the `Connection` since the event is dispatched in the context of an indexer. If we would use the Data Abstraction Layer \(DAL\) for writing changes to the database, the indexer would be triggered again, because it listens for `EntityWrittenContainerEvent` events. This would lead to an infinite loop. Using the `Connection` directly prevents the DAL from dispatching entity written events.
```
Also link to ### Use DAL functionalities in the indexer

pweyck (changes requested):

guides/plugins/plugins/framework/data-handling/add-data-indexer.md
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    $context = $event->getContext();
+    $context->addExtension(EntityIndexerRegistry::DISABLE_INDEXING, new ArrayEntity());
+
+    return new EntityIndexingMessage(array_values($updates), null, $context);
+}
+```
+
+## Index data using existing events
```
pweyck, line 178: Link https://github.com/shopware/platform/blob/trunk/adr/2021-05-14-when-to-use-plain-sql-or-dal.md somewhere

PaddyS (changes requested):

PR #298, opened 2021-06-02
NEXT-12626 - Add custom media type guide

PaddyS:


PaddyS:
@JoshuaBehrens Nope, totally agree!
Now I can't "unread" it and always have to think about a DAL media extension 😂

PR #300, opened 2021-06-10
Add argument to DOM binding

despokd:
I figured out that DOM binding only works if  `data-example-plugin` has a value like `true` (so not `false` or `undefined`).

### DIFF
--- a/guides/plugins/plugins/storefront/add-custom-javascript.md
+++ b/guides/plugins/plugins/storefront/add-custom-javascript.md
@@ -112,7 +112,7 @@ A lot of text, here is the respective example:
 {% block base_main_inner %}
     {{ parent() }}
 
-    <template data-example-plugin></template>
+    <template data-example-plugin="true"></template>
 {% endblock %}
 ```
 {% endcode %}


PaddyS:
Hey @despokd,

thank you for your contribution! 🎉 
I will try to take a look at it and merge it right afterwards, once I can confirm the issue. :-)

Thank you very much!

PaddyS:
Hey @despokd,

so I had just had a look at it. :-)

Unfortunately, I couldn't really reproduce the issue you're reporting.
I just did exactly what the guide said and it's instantly working.

However, I can think of it being an issue depending on the browser or OS, probably might be a difference in the handling of the data-attributes.

Can you ellaborate a little bit on which browser and OS this issue occurred for you? :-)
I've tried this with Firefox and Google Chrome on MacOS.

despokd:
Hi @PaddyS,

I tested it again, without an argument, and it worked. I guess I forget to try `php bin/console cache:clear` after `./psh.phar storefront:build`.

For completeness:
OS: Windows 10 Pro 21H1 with Docker
SW: 6.4.0 with dockware:cont
Browser: Edge Chromium 91

PR #304, opened 2021-06-16
Adjust note regarding @internal annotation

philipgatzka:
This clarifies, that usage of the `@internal` annotation is allowed *without* a feature flag as well.

### DIFF
--- a/resources/guidelines/code/backward-compatibility.md
+++ b/resources/guidelines/code/backward-compatibility.md
@@ -48,7 +48,7 @@ The `@major-deprecated` annotation is used for breaking code which has to stay b
  */
 ```
 
-The `@internal` annotation is used for new introduced code, which is not yet released or hidden behind a feature flag. This ensures, that it will not be treated as public API until the corresponding feature is released and makes it possible to change the code in any way until the final release. Always add the name of the corresponding feature flag to the annotation, so that it will not be forgotten, when the corresponding feature is released.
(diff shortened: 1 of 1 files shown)


tobiasberge:
@ssltg @philipgatzka Is this really the case that `@internal` is supposed to be converted into "public API" in the future? I always thought somehow that this a permanent thing like "This function is `@internal` and not supposed to be overwritten or extended"

tobiasberge:
@philipgatzka Yes I think it sounds better than before but this sentence sounds a bit like that it is only for new code before a release and will always be removed after a release.

```
the `@internal` annotation is used for newly introduced code, which is not yet released.
```
Maybe we could add another sentence which makes clear that we can use this to mark stuff as internal in general

ssltg:
> 
> 
> @philipgatzka Yes I think it sounds better than before but this sentence sounds a bit like that it is only for new code before a release and will always be removed after a release.
> 
> ```
> the `@internal` annotation is used for newly introduced code, which is not yet released.
> ```
> 
> Maybe we could add another sentence which makes clear that we can use this to mark stuff as internal in general

For me the preceding "In combination with..." makes this already clear. If you think this is still ambiguous, how would you change the sentence?

tobiasberge:
Maybe something extra like: "In addition you can use `@internal` to mark parts of code which are not supposed to be extended or overwritten in general and are not supposed to be public API in the future".

But if it's also clear for you it may just be me. :D

tobiasberge (approved):

PR #317, opened 2021-06-24
replaces return; in foreach with continue;

riconeitzel:
I might be wrong but this should be continue, isn't it?


PaddyS:
Hey @riconeitzel,

I totally agree here. Thanks for your PR! 🎉

PR #319, opened 2021-06-27
fixes Flags class Names

riconeitzel:


PaddyS:
Thank you once again @riconeitzel 🎉

PR #321, opened 2021-06-29
move main.js entry to the right place

dominikmank:
Changed the tree for main.js to the one as described in the text before "In the case of JavaScript, you have your `main.js` as entry point which has to be located the `src/Resources/app/storefront/src/` directory:
"


PaddyS:
Hey @dominikmank,

thank you very much for your contribution! 💙

PR #322, opened 2021-06-29
added config & theme_config twig functions

dominikmank:


PaddyS:
Awesome, thank you @dominikmank 🎉

PR #323, opened 2021-06-29
use config()

dominikmank:
...for getting the config in twig templates


guides/plugins/apps/configuration.md
```diff
@@ -43,6 +43,6 @@ Keep in mind that your app needs the `system_config:read` permission to access t
 Inside twig templates you can use the `config` property of the global `shopware` object to read your configuration. An example twig template could look like this:
```
tinect, line 43: This sentence should be adjusted, too 🤔

tinect (changes requested):

guides/plugins/apps/configuration.md
dominikmank: never enough :D /done

PR #324, opened 2021-06-29
Fix config access

mitelg:
config access is wrong (since 6.4)


mitelg:
duplicates https://github.com/shopware/docs/pull/323

PR #325, opened 2021-06-30
Add hints how to make your app look like the admin

JoshuaBehrens:
I published an app and wanted it to be nice. It was like the Shopware 5 lightweight modules. No style or tooling given. Since the pre-compiled packages are available it is now super easy to match the styles :) I tell you how I was able to pull it off

### DIFF
--- a/guides/plugins/apps/administration/add-custom-modules.md
+++ b/guides/plugins/apps/administration/add-custom-modules.md
@@ -141,3 +141,10 @@ To avoid mixing other modules with your main module, we decided to separate the 
 
 This feature is not compatible with themes as they will always open the theme config by default.
 
+## Admin design compatibility
+
+As your module page is integrated as an iframe you are not able to use the stylesheet and javascript out of the box.
+Having the stylesheets that are used in the administration can be beneficial for the app module to seamlessly integrate into the administration.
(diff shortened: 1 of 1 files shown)


guides/plugins/apps/administration/add-custom-modules.md
```diff
@@ -141,3 +141,10 @@ To avoid mixing other modules with your main module, we decided to separate the
 
 This feature is not compatible with themes as they will always open the theme config by default.
 
+## Admin design compatibility
```
keulinho, line 144: We are aware of the difficulties of getting the look&feel of your iframes to match the rest of the administratrion, we have a solution in mind, but are not sure when we can deliver on that. For the meantime i am fine with this creative approach. @Christian-Rades what do you think?

Christian-Rades: Agreed, I think it's a good Idea :+1:

JoshuaBehrens: "Creative approach" :) I don't see where programming isn't creative

keulinho (approved):

keulinho:
Thanks for your contribution @JoshuaBehrens :tada: :blue_heart:

PR #326, opened 2021-06-30
Update manifest schema reference

JoshuaBehrens:
When you copy the reference you'll get lots of xsd related issues (depending on the IDE/editor). When we use the latest reference we are good to go, aren't we?

### DIFF
--- a/resources/references/app-reference/manifest-reference.md
+++ b/resources/references/app-reference/manifest-reference.md
@@ -3,7 +3,7 @@
 {% code title="manifest.xml" %}
 ```markup
 <?xml version="1.0" encoding="UTF-8"?>
-<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/app-system/0.1.0/src/Core/Content/App/Manifest/Schema/manifest-1.0.xsd">
+<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/app-system/0.2.0/src/Core/Content/App/Manifest/Schema/manifest-1.0.xsd">
     <meta>
         <!-- This is the element for the technical name of your app and must equal the name of the folder your app is contained in -->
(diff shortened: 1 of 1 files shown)


resources/references/app-reference/manifest-reference.md
```diff
@@ -3,7 +3,7 @@
 {% code title="manifest.xml" %}
 ```markup
 <?xml version="1.0" encoding="UTF-8"?>
-<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/app-system/0.1.0/src/Core/Content/App/Manifest/Schema/manifest-1.0.xsd">
+<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/app-system/0.2.0/src/Core/Content/App/Manifest/Schema/manifest-1.0.xsd">
```
keulinho, line 6: definetly an error that we still reference the app-system plugin, as the app-system now is integrated into the platform

so we should use the current schema file from the platform repo, please use the `https://raw.githubusercontent.com/shopware/platform/trunk/src/Core/Framework/App/Manifest/Schema/manifest-1.0.xsd` url

```suggestion
<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/platform/trunk/src/Core/Framework/App/Manifest/Schema/manifest-1.0.xsd">
```

keulinho (changes requested):
@JoshuaBehrens thanks for pointing out the outdated reference, but please update it to the current platform schema, as that link should not get outdated anytime sonn.

resources/references/app-reference/manifest-reference.md
JoshuaBehrens: Using trunk is a good one so you are not missing to update again :)

JoshuaBehrens:
@keulinho do we need a squash for that?

keulinho (approved):

keulinho:
I don't think a squash is needed here

PR #330, opened 2021-07-05
Generate correct json in html attributes

JoshuaBehrens:
Three things are fixed here:

* We are not handcrafting JSON anymore
* We are not using double quotes for html attributes and double quotes for json
* This is not how it is used anywhere in the shopware code (checked in 6.4.1.2)


PaddyS:
Just double-checked - seems to have changed indeed. :-)

Thank you very much! 🎉

PR #331, opened 2021-07-05
Be more clear about mysql version 8

bjoern-flagbit:
The problem is that a lot of people think mysql 8 version in TOTAL is not compatible.

### DIFF
--- a/guides/installation/overview.md
+++ b/guides/installation/overview.md
@@ -74,7 +74,7 @@ PHP
 SQL
 
 * MySQL 5.7.21 or higher
-  * MySQL 8.0.20 is not compatible
+  * MySQL 8.0.20 is not compatible (only this specific version, 8.0.23 for example is working)
 * MariaDB 10.3.22 or higher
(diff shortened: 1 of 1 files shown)


guides/installation/overview.md
```diff
@@ -74,7 +74,7 @@ PHP
 SQL
 
 * MySQL 5.7.21 or higher
-  * MySQL 8.0.20 is not compatible
+  * MySQL 8.0.20 is not compatible (only this specific version, 8.0.23 for example is working)
```
PaddyS, line 77: I'm not entirely sure why we would need a change here to be honest, as it literally only mentions a single version and not MySQL as a whole.
However, I guess I'd rather have it like this. What do you think?

```suggestion
  * Only MySQL 8.0.20 in specific is not compatible
```

PaddyS (changes requested):

guides/installation/overview.md
bjoern-flagbit: Fine with that :+1: 
I heared already 5 peoples that thinks that mysql 8 in total is not supported.
So I am totaly sure this change is needed. Thx.

PR #334, opened 2021-07-09
NEXT-14361 - Smartbar buttons open modal with iframe

nguyenngon95:


guides/plugins/apps/administration/add-custom-action-button.md
```diff
@@ -108,8 +108,23 @@ To reload the data in the user's current page you can use the `reload` action ty
 
 ```
 
-* `actionType`: The type of action the app want to be triggered, including `notification`, `reload`, `openNewTab`
+To open a modal with the embedded link in the iframe, you can use the `openModal` action type. You need to pass the url that should be opened as the `iframeUrl` property and the `size` property inside the payload.
```
keulinho, line 111: Please indicate that this action type is only supported from version 6.4.3.0 onward

keulinho (changes requested):

guides/plugins/apps/administration/add-custom-action-button.md
nguyenngon95: @keulinho i think it has the hint in line 74 for the whole type response

keulinho (approved):

PR #335, opened 2021-07-13
Make the scope of the decoration easier to understand

Schrank:
Add a note to make clear, what this page is about. It is NOT about the service or abstract class, but about the decoration.

### DIFF
--- a/guides/plugins/plugins/plugin-fundamentals/adjusting-service.md
+++ b/guides/plugins/plugins/plugin-fundamentals/adjusting-service.md
@@ -43,7 +43,7 @@ xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/sc
 Now we have to define an abstract class because it's more beautiful and not so strict like interfaces. With an abstract class we can add new functions easier, you can read more about this at the end of this article. The abstract class has to include an abstract function called `getDecorated()` which has the return type of our instance.
 
 {% hint style="info" %}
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/plugin-fundamentals/adjusting-service.md
```diff
@@ -42,6 +42,11 @@ xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/sc
 
 Now we have to define an abstract class because it's more beautiful and not so strict like interfaces. With an abstract class we can add new functions easier, you can read more about this at the end of this article. The abstract class has to include an abstract function called `getDecorated()` which has the return type of our instance.
 
+{% hint style="info" %}
+    To avoid misunderstandings: The abstract service class and the implementation of it is not part of the decoration process itself and comes most of the times either from the shopware core or from a plugin you want to extend. They are added here to have an example to decorate.
```
PaddyS, line 46:
```suggestion
To avoid misunderstandings: The abstract service class and the implementation of it is not part of the decoration process itself and most of the times comes either from the Shopware core or from a plugin you want to extend. They are added here to have an example to decorate.
```

PaddyS (changes requested):
What do you think? :-)

Schrank:
@PaddyS pretty cool feature didn't know this is possible, thanks for the better wording!

PaddyS:
Thank you very much, @Schrank 🎉

PR #340, opened 2021-07-18
fix db table name in scheduled task docs

rkusa:
The table is called `scheduled_task` at least in my installation, so I'd assume that this is a typo 😊


shyim:
Thanks!

PR #343, opened 2021-07-21
missing adjustment after swap of dev documentation

ufcyg:
the former bundle tutorial is now displayed in a more generalized manner. at "ExampleCollection" the namespace still contained "Bundle" instead of "Example"


PaddyS:
Thank you very much, @ufcyg 💙

PR #348, opened 2021-07-26
Update example with new QuantityPriceDefinition signature

rickschippers:
The signature of the QuantityPriceDefinition was changed. Updated the code example for the new version.


PaddyS:
Thank you very much @rickschippers 💙

PR #351, opened 2021-07-28
NEXT-15824 - Allow apps to provide error message when installation fails

Tamvt:
We should allow apps to provide error message when installation fails

### DIFF
--- a/guides/plugins/apps/app-base-guide.md
+++ b/guides/plugins/apps/app-base-guide.md
@@ -131,6 +131,14 @@ $signature = hash_hmac('sha256', $queryString, $appSecret);
 
 ### Registration Response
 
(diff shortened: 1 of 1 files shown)


guides/plugins/apps/app-base-guide.md
```diff
@@ -131,6 +131,14 @@ $signature = hash_hmac('sha256', $queryString, $appSecret);
 
 ### Registration Response
 
+There may be valid cases where the app installation fails, because the domain is blocked, or some other prerequisite in that shop is not met, in which case you can return the message error as follows
+```javascript
+{
+  "error": "Shop url is not met"
```
Christian-Rades, line 137: Meeting a URL is a phrase I don't know 🤔
Do you mean reachable or it didn't meet validation criteria? In the latter case I would call it "the shop URL is invalid"

Christian-Rades (changes requested):

PR #353, opened 2021-07-28
use method instead of property in code example

riconeitzel:


PaddyS:
Thank you very much @riconeitzel 💙

PR #354, opened 2021-07-28
Update add-custom-component.md

stephfuchs:
This seems to be a mix-up in the administration path for the component, because the import path isn't matching with the paths for the custom component. If my changes are incorrect, than please change the import path for the "hello-world/index.js" in line 49.

Best wishes


PaddyS:
I do think you're right, @stephfuchs.
Thank you very much! 🎉

PR #360, opened 2021-08-02
Ensure topicality of varnish article

leichteckig:


guides/hosting/infrastructure/reverse-http-cache.md
```diff
@@ -13,6 +13,8 @@ In this guide, we will use Varnish as example for an http cache.
 
 ### The example Setup with Varnish
```
leichteckig, line 14: As the whole article is about this example setup, would it be better to remove this article from version 6.3.0 altogether?

elkmod (approved):

PR #362, opened 2021-08-03
Change CLI command for message consume

aherking:
Deprecated command bin/console messenger:consume-messages replaced with regular command bin/console messenger:consume


elkmod:
Thanks for your contribution @aherking

PR #364, opened 2021-08-05
replaces $logger with $cacheInvalidator in example

riconeitzel:
the last code block was using $logger for the cacheInvalidator instance …


PaddyS:
🤦 Thank you very much! 💙

PR #367, opened 2021-08-07
How to rebuild administration in add-new-tab.md

wirduzen-felix:
I noticed that the only page in the docs that ever mentions that you need to rebuild administration after you change your main.js is add-custom-field.md. This was very confusing to me and I think it should be added to pretty much every page in the docs that makes changes to main.js.

### DIFF
--- a/guides/plugins/plugins/administration/add-new-tab.md
+++ b/guides/plugins/plugins/administration/add-new-tab.md
@@ -132,6 +132,22 @@ This is an example of what your `main.js` should look like in order to load your
 ```javascript
 import './page/sw-product-detail';
 ```
+{% hint style="info" %}
+Don't forget to rebuild the administration after making changes to your `main.js`.
+{% tabs %}
+{% tab title="Development template" %}
+```bash
+./psh.phar administration:build
+```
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/administration/add-new-tab.md
```diff
@@ -132,6 +132,22 @@ This is an example of what your `main.js` should look like in order to load your
 ```javascript
 import './page/sw-product-detail';
 ```
+{% hint style="info" %}
+Don't forget to rebuild the administration after making changes to your `main.js`.
```
PaddyS, line 136:
```suggestion
Don't forget to rebuild the administration after applying changes to your `main.js`.
```

PaddyS (changes requested):

PaddyS:
Thank you very much, @wirduzen-felix! 🎉 

Just some minor wording change, and then it's perfectly fine. Appreciate that you also used the tabs solution for the different setups. 💙

PaddyS:
Thank you very much, @wirduzen-felix 🎉

PR #368, opened 2021-08-09
Updated the link to dockware default credentials

calcifer1337:
The original link ist broken


PaddyS:
Thank you very much, @calcifer1337 🎉

PR #372, opened 2021-08-12
Make it explicit WHEN plugin init is running

Schrank:
Be more specific when init() runs


Schrank:
@PaddyS (Sorry if you are the wrong one) - I think the link checker has a bug :-) And should ignore "

elkmod:
Thanks for the contribution @Schrank :)

PR #379, opened 2021-08-20
NEXT-16770 - Update App CMS blocks

fschmtt:
Update docs on adding custom CMS blocks with an App for the upcoming release.

### DIFF
--- a/guides/plugins/apps/content/cms/add-custom-cms-blocks.md
+++ b/guides/plugins/apps/content/cms/add-custom-cms-blocks.md
@@ -12,6 +12,10 @@ If you're not yet familiar with custom CMS blocks you might want to head over to
 
 {% page-ref page="../../../plugins/content/cms/add-cms-block.md" %}
 
+## Prerequisites
+
+This guide is based on our [App Base Guide](../../app-base-guide.md) and assumes you have already set up an app.
+
 ## Overview
 
 Adding custom CMS blocks from an app works a bit differently than [adding them from a plugin](../../../plugins/content/cms/add-cms-block.md).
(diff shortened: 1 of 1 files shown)


SebastianFranze (approved):

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
```diff
@@ -20,36 +20,38 @@ The basic directory structure looks as follows:
...
+│   │           └── cms-block-swag-image-text-reversed-component.html.twig
 │   └── cms.xml
 └── manifest.xml
 ```
 
 Each CMS block defined within your `cms.xml` must have a directory matching the block's name in `Resources/cms/blocks/`.
-In those directories you shape your blocks by supplying a `preview.html` containing the template used for displaying a preview.
-Styling the preview is possible from the `styles.css`.
+In those directories you shape your blocks for the CMS editor in the Administration by supplying a `preview.html` containing the template used for displaying a preview.
+Styling the preview in the sidebar and the component on the CMS stage is possible from the `styles.css`.
```
xPand4B, line 43: `CMS stage` sound a bit weird to me, what about `CMS page`?

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
```diff
@@ -20,36 +20,38 @@ The basic directory structure looks as follows:
...
 
 {% hint type=info %}
 Due to technical limitations it's not possible to use templating engines \(like Twig\) or preprocessors \(like Sass\) for rendering and styling the preview.
 {% endhint %}
 
+The Storefront representations reside in the `Resources/views/storefront/block/` directory.
```
xPand4B, line 49:
```suggestion
The Storefront representations of our new block(s) reside in the `Resources/views/storefront/block/` directory.
```

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
```diff
@@ -20,36 +20,38 @@ The basic directory structure looks as follows:
...
+
 ## Defining blocks
 
 As already mentioned above and similarly to an app's `manifest.xml` CMS blocks also require some definition which is done in a `cms.xml`.
```
xPand4B, line 53:
```suggestion
As already mentioned above and similarly to an app's `manifest.xml`, CMS blocks also require some definition which is done in a `cms.xml`.
```
This part seems a bit out-of-order for me as well.

xPand4B (changes requested):
I'm not wuite sure what's expected, but maybe we could add a screenshot of the result or sth like this?

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
fschmtt: The CMS stage is the part where you drag-and-drop your blocks into.

But you're kind of right it's not used anywhere else in the docs and might be somewhat confusing.

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
fschmtt: To stay consistent I'll use `your` instead of `our` 😄

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
fschmtt: Where would you expect it? Or is the information redundant to you?

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
fschmtt: I'll rename it to `CMS editor` instead of `CMS stage` - I think this is more comprehensible.

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
xPand4B: It's not redundant but you start by talking about the `cms.xml` and then continue with this part about the `manifest.xml`...I was asking myself where the `manifest.xml` even was mentioned to beginn with. 

So maybe we could write a short part about the `manifest.xml` or link to the other site where it is mentioned.

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
xPand4B: CMS panel? 🤔

guides/plugins/apps/content/cms/add-custom-cms-blocks.md
fschmtt: I don't want to cover how to set up a `manifest.xml` in this article, I just state that "just as you already know, like for the app itself, provide the configuration as XML".

If one has never set up a `manifest.xml` previously, they won't get to the point where they could actually add a CMS block 😄

xPand4B: Alright :) But pelase add the comma 😄

SebastianFranze (approved):

xPand4B (approved):

PR #380, opened 2021-08-22
fix the command name

waldicom:


PaddyS:
Awesome! Thank you, @waldicom! 🎉

PR #381, opened 2021-08-23
Fix wrong filename

Tschucki:
We registered the cms-block with:
name: 'my-image-text-reversed'.

Accordingly, the block must be named 'cms-block-my-image-text-reversed.html.twig' instead of 'cms-block-image-text-reversed.html.twig'.
Otherwise the Block won't get rendered.

### DIFF
--- a/guides/plugins/plugins/content/cms/add-cms-block.md
+++ b/guides/plugins/plugins/content/cms/add-cms-block.md
@@ -269,7 +269,7 @@ A block's storefront representation is always expected in the directory [platfor
 
 So go ahead and re-create that structure in your plugin: `<plugin root>/src/Resources/views/storefront/block/`
 
-In there create a new twig template named after your block, so `cms-block-image-text-reversed.html.twig` that is.
+In there create a new twig template named after your block, so `cms-block-my-image-text-reversed.html.twig` that is.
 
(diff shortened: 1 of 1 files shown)


PaddyS:
Hey @Tschucki,

thank you for your pull request! 💙 
However, I might need to do some double-checking here.

Afaik, the name of the `.html.twig` should be equal to the component of the block being used, which in this example is `sw-cms-block-image-text-reversed` and therefore perfectly fits.

Did you maybe also change the component?

PaddyS:
Ping @Tschucki

thenostromo:
I agree with @Tschucki 
Only in this way it will work.

Filename shoud be 'cms-block-my-image-text-reversed.html.twig'.

PaddyS:
@thenostromo Maybe you can shed some light here?

Doesn't it work for you as well?
The issue I'm having, is: I just created a new plugin using the guide and exactly the same code from the guide - and it works flawlessly.

Can you maybe elaborate [on this question](https://github.com/shopware/docs/pull/381#issuecomment-905385355)?

elkmod:
We are going to close this pull request due to inactivity and not being able to reproduce the issue. Please reopen, if you feel like this should be fixed.

PR #383, opened 2021-08-25
Fix class name of extended service

dallyger:
The code example above does reference `AbstractExampleRoute` and not `AbstractExampleService` as described below.


PaddyS:
You're absolutely right here. Thank you @dallyger! 🎉

PR #384, opened 2021-08-27
Add hint of required composer key to plugin base guide

dallyger:
This was stated already under the 'Add plugin dependencies' guide,
but really should be added to the plugin introduction too.


PaddyS:
Thank you a lot @dallyger!

PR #386, opened 2021-08-30
Fix path in example

bpeschlaudert:
The described example is creating a storefront plugin hence the administration path is incorrect.
The [linked video](https://www.youtube.com/watch?v=wfBuWdff35c) also uses the `storefront` namespace.


PaddyS:
Thank you very much @bpeschlaudert! 💙

PR #387, opened 2021-08-31
Update from-scratch.md

Xarno:
For Apache 2.4 „Require all granted“ is needed in <Directory> elements.


PaddyS:
Oh boy, nice catch! Just looked it up and you seem to be right here. Thanks a lot @Xarno 🎉

PR #388, opened 2021-09-03
Add a missing link for the create migration command to the migrations…

J-Rahe:
The Page only says that there is an command for creating an migration, but does not mention which one.
This PR adds an link to the database-migrations guide.


J-Rahe:
There is already an link to the guide in the next section, but it seems confusing to me to just write there is an command and then not name it directly.

PaddyS:
Thank you, @J-Rahe! Totally agree here!

PR #389, opened 2021-09-06
Add note about plugin styles import for StorefrontBootstrap

s-diez:
`@StorefrontBootstrap` does not include `@Plugins` in contrast to `@Storefront`. Added it into example and added a note.


JoshuaBehrens (approved):

PaddyS:
Nice addition, @s-diez! Thank you very much! 🎉

PR #391, opened 2021-09-08
NEXT-15838 - Update documents for request header context lang and locale

nam4am:
update document for shopware context language and locale user

### DIFF
--- a/guides/plugins/apps/app-base-guide.md
+++ b/guides/plugins/apps/app-base-guide.md
@@ -104,6 +104,7 @@ GET https://my.example.com/registration?shop-id=KIPf0Fz6BUkN&shop-url=http%3A%2F
 
 {% hint style="info" %}
 Starting from Shopware version 6.4.1.0, the current shopware version will be sent as a `sw-version` header.
+Starting from Shopware version 6.4.5.0, the current language id of the shopware context will be sent as a  `sw-context-language` header , and the locale of the user or locale of the context language is available under the `sw-user-language` header.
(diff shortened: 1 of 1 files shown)


guides/plugins/apps/app-base-guide.md
```diff
@@ -104,6 +104,7 @@ GET https://my.example.com/registration?shop-id=KIPf0Fz6BUkN&shop-url=http%3A%2F
 
 {% hint style="info" %}
 Starting from Shopware version 6.4.1.0, the current shopware version will be sent as a `sw-version` header.
+Starting from Shopware version 6.4.5.0, the current shopware context language-id `sw-context-language`, and `sw-user-language` represent for the locale of the user or locale of the context language, will be sent to every request header.
```
keulinho, line 107:
```suggestion
Starting from Shopware version 6.4.5.0, the current language id of the shopware context will be sent as a  `sw-context-language` header , and the locale of the user or locale of the context language is available under the `sw-user-language`-header.
```

please adjust the other occurences accordingly

JoshuaBehrens:
It is cool that we have this feature now. Can we link at some other place where one can see how it is used? It is not obvious what to do with that new values without knowing a context can have a language chain.

keulinho:
@JoshuaBehrens i've merged this PR as is right now, feel free open another PR and add some more information to make things clearer

JoshuaBehrens:
I put it on my bucket list. Thanks for the ping.

PR #394, opened 2021-09-14
docs(data-handling): fix method name referenced in text

evolkmann:
The correct method name `getDefinitionClass()` is mentioned in the code block below.


PaddyS:
Awesome, thank you @evolkmann! 🎉

PR #395, opened 2021-09-14
RepositoryIterator: Add note about determinism

Josniii:
See https://github.com/shopware/platform/pull/2070

Since this issue seems to have shown up in the core codebase, I figured it would be good to have a note in the documentation on the importance of having deterministic sorting when using the RepositoryIterator. Otherwise developers risk using the abstraction with a bad sorting which results in unwanted behaviour.

If the issue I'm trying to describe is unclear or you'd like the wording changed please let me know.

### DIFF
--- a/guides/plugins/plugins/framework/data-handling/reading-data.md
+++ b/guides/plugins/plugins/framework/data-handling/reading-data.md
@@ -364,6 +364,21 @@ public function readData(Context $context): void
 
 In this example, you'd get a batch of 500 products with each iteration of the `while` loop. This way, `$result` will not cause a "memory exhausted" error and you can handle huge amounts of data this way.
 
+One small caveat to be aware of: When using the `RepositoryIterator`, make sure that the `Criteria` uses a sorting which is deterministic. 
+
+Put differently, you must ensure that your sorting means that there's only one correct way to order your results, otherwise different batches might decide to sort the entities differently in the database. That would mean you risk getting the same entity in several batches (and having entities that won't be iterated at all).
+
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
This is an important hint. I think you have this already on definitions that have an autoIncrement column, right?

Josniii:
I'm not sure I understand what you're asking, but ideally you'd always specify a sort order when using something like this where the underlying query behind the abstraction ends up doing OFFSET/LIMIT , because technically you can't rely on MySQL always giving results in the same order if no ORDER BY clause is specified (or the ORDER BY clause is ambiguous), like the Manufacturer Number example i put in the PR. 

Normally with InnoDB tables you'd get things ordered by primary key if no sorting was specified, but depending on indexes used for fetching etc. you may get different results depending on what MySQL decides is optimal. So while in 99% of cases you won't see any issues, ideally you'd probably always have a good sorting clause in the criteria used for batch loading.

Auto-increment columns are good for resolving this ambiguity because they're usually unique and indexed, and so sorting on them is deterministic and relatively fast. As is the ID for the same reasons, which is why I used the ID for this example. If that's what you're asking.

JoshuaBehrens:
@Josniii I am totally on your side. Basically your text should be the hint in the docs :D . Sorry in my head I was here `\Shopware\Core\Framework\DataAbstractionLayer\Dbal\Common\IteratorFactory::createIterator`
https://github.com/shopware/platform/blob/d6c853ca3b6b167d61d70e6329f4fcface5c5245/src/Core/Framework/DataAbstractionLayer/Dbal/Common/IteratorFactory.php#L27-L37

and thought I was in the RepositoryIterataor. But I was not. So my thought: "hey, you might not need this as it automatically falls back to autoIncrement" was based on the other iterator 😊 sorry

PaddyS:
Hey @Josniii, I totally love it! Thank you very much!

PR #396, opened 2021-09-17
Update dependency-injection.md

HashtagAssist:
Constructors in object-oriented languages do not have return types.
https://bugs.php.net/bug.php?id=75263

And there is a wrong path specified in "guides/plugins/plugins/storefront/add-custom-controller.md"


shyim:
Thanks!

PR #408, opened 2021-09-29
SWAG-181738 - Clarify rule variable visibility

marcelbrode:
Got already 3 customers in support tickets, who made the variable private instead of protected, so I decided to update the docs according to that :smile:


PaddyS:
Thanks a lot! 🎉

PR #411, opened 2021-10-08
Escape JSON to ensure correct HTML parsing

JoshuaBehrens:
When you use the code you will eventually end up with wrong HTML :( When you pass in null it will also end up bad in the option merging later on. Why shall it be null? It can be because you propagate to allow others to change that value. Makes the code a bit more bullet proof


JoshuaBehrens:
Thank you for merging :)

PR #447, opened 2021-12-03
Remove the exact PHP Version from the response for more security

Donderda:
Remove the exact PHP Version from the response for more security


Christian-Rades:
Hi @Donderda, thanks for Contributing!

PR #452, opened 2021-12-10
Fix for misspell of "their"

sylphlv:
I think it was supposed to be "their" instead of "there". 
"Identified by there URL" doesn't make sense to me.


elkmod:
Thanks for your contribution @sylphlv 💙

PR #453, opened 2021-12-13
Varnish config: Administration not compatible

Josniii:
The sample Varnish config doesn't handle PATCH requests which causes issues when the administration passes requests through Varnish.

This PR adds this.


elkmod:
Thanks for your contribution @Josniii

PR #462, opened 2022-01-04
Add a guide for creating own twig functions

Tschucki:
During the theme development for a shop, I often got to the point where I wished there was a Twig Function for it.

For example:
Check if a product is within a given category.
Find all available variants for a product
Flatten category tree
Find sibling categories
Find out how many products are in a category
etc.

Unfortunately I couldn't find anything about this in the documentation.

That's why I think the guide should be included in the documentation.

The attached commit contains that guide. (kinda)

### DIFF
--- a/guides/plugins/plugins/storefront/add-custom-twig-function.md
+++ b/guides/plugins/plugins/storefront/add-custom-twig-function.md
@@ -1,0 +1,84 @@
+# Add custom Twig Functions
+
+## Overview
+
+Imagine you want to call a PHP script from the Twig template during the theme development to retrieve data from the database or to check whether a category is within the current path. For these cases we can create our own twig functions.
+
+In our example we will pass a category ID to the Twig Function and return the category object.
+
+## Prerequisites
+
+In order to create your own Twig Function for your plugin, you first need a plugin as base. Therefore, you can refer to the [Plugin Base Guide](../plugin-base-guide.md).
+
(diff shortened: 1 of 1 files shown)


ssltg:
Hi @Tschucki ,
thank you for your contribution. Your documentation is well written 👍 
Two things:
1. The documentation should also include an example how the new function is used in a template (it is obvious, but better have it, than someone has to ask)
2. More serious: The chosen example is not the best to provide. Of course one can do such a twig function, but it is not recommended to add DAL Calls inside twig functions because of performance and cache reliability. Can you change the example to something that works on the provided Page data in a template?

If you currently use twig functions to get more data you are missing, you should follow this guide instead: https://developer.shopware.com/docs/guides/plugins/plugins/storefront/add-data-to-storefront-page

regards,
Stefan

ssltg (commented):
...

ssltg (changes requested):
See comment https://github.com/shopware/docs/pull/462#issuecomment-1005718519

elkmod:
@Tschucki - could you please provide a different example, which does adhere to the "guidelines" @ssltg mentioned above? E.g. a "hash" method, which returns hashed string of a certain input string?

elkmod:
We're closing this PR for inactivity. It remained open, because it doesn't adhere to our guidelines yet. Feel free to reopen, once those issues are resolved.

PR #463, opened 2022-01-07
added parent_id field for reference

SKolev1989:
adjusted migration for parent_id field needed by ParentFkField;
adjusted default values for $parent and $children -> if you don't set them default to null, you get an error:
"$parent must not be accessed before initialization"


shyim (approved):
lgtm

PR #467, opened 2022-01-13
Update reverse-http-cache.md to be better linkable

tdgroot:
Also add documentation about disabling the verification headers


shyim:
Thanks

PR #469, opened 2022-01-17
Moved one section to not be indented

Kudze:
In webpage is hard to scan when this paragraph, looks like it belongs to update section, but in reality it does not.


Kudze:
Intended result is this:
image

elkmod:
Thanks for your contribution @Kudze

PR #473, opened 2022-01-18
Update theme-inheritance-configuration.md

andriichuk:


elkmod:
Thanks for the fix @andriichuk

PR #475, opened 2022-01-23
Add info about default cypress.json

Schrank:
It took me a while to realize, that I need a env.local set, to make `cy.login();` work. Next problem was 

    cy.clickMainMenuItem({
        targetPath: '#/sw/product/index',
        mainMenuId: 'sw-catalogue',
        subMenuId: 'sw-product'
    });

which does NOT work on the default 1000x660px viewport.


leichteckig (approved):

leichteckig:
Ah, good catch! 🙏  Thank you @Schrank for your contribution. 💙  That's approval. ✅

PR #477, opened 2022-01-28
Add example for Redis message queue

shyim:
Preinstalling PR https://gitlab.shopware.com/shopware/6/product/platform/-/merge_requests/7340

### DIFF
--- a/guides/hosting/infrastructure/message-queue.md
+++ b/guides/hosting/infrastructure/message-queue.md
@@ -110,7 +110,7 @@ framework:
 
 ### Transport: RabbitMQ example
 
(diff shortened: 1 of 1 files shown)


keulinho (approved):

guides/hosting/infrastructure/message-queue.md
```diff
@@ -133,3 +133,32 @@ framework:
 ```
 
 Notice that `shopware-queue` is the name of the previously created queue in RabbitMQ. Also `rabbitmq` matches the name of the new transport in the previously created `enqueue.yaml`.
+
+### Transport: Redis example
+
+In the following I assume that Redis is installed and started. Since Shopware 6.4.9.0 is the Redis transport preinstalled. In previous versions this must be installed using `composer require enqueue/redis`. If the Redis PHP extension is not installed, you need also to install Predis using `composer require predis/predis`. The only thing left is to tell Shopware about the new transport. Therefore we edit/create the configuration file `enqueue.yaml` with the following content:
```
soebbing, line 139:
```suggestion
In the following I assume that Redis is installed and started. Since Shopware 6.4.9.0, the Redis transport is preinstalled. In previous versions this had to be installed using `composer require enqueue/redis`. If the Redis PHP extension is not installed, you also need to install Predis using `composer require predis/predis`. The only thing left is to tell Shopware about the new transport. Therefore, we edit/create the configuration file `enqueue.yaml` with the following content:
```

PR #483, opened 2022-01-31
Fix grammar for transactions

kevinvhengst:


elkmod:
Thanks for the fix @kevinvhengst! :)

PR #485, opened 2022-02-10
fixed typo in first line of the administration - add menu entry guide

pajonas:
Found a little typo in the docs. Here's the fix.


elkmod:
Thanks for the fix @pajonas, the issue was just removed.

PR #486, opened 2022-02-10
Restore codes removed in commit 1a34e71

roma84:
When reading [Add SCSS Styling and JavaScript to a Theme](https://developer.shopware.com/docs/guides/plugins/themes/add-css-js-to-theme) I noticed code blocks are broken. 
The codes were mistakenly removed in this commit https://github.com/shopware/docs/commit/1a34e7180e9c187a478755cb474f40a29465383c.


elkmod:
Thanks for fixing those @roma84 !

PR #490, opened 2022-02-15
fix: typo add menu entry

Zerotask:


elkmod:
Thanks for your contribution @Zerotask !

PR #492, opened 2022-02-18
Fix further reading links

JoshuaBehrens:
image


elkmod:
Thank you @JoshuaBehrens :)

Issue #500, opened 2022-03-01
Amend sample `phpunit.xml.dist` for plugins to ignore Symfony deprecations


svlauer:
The doc page on how to run unit tests for plugins includes [a section with a recommended PHPUnit configuration file](https://developer.shopware.com/docs/guides/plugins/plugins/testing/php-unit#phpunit-configuration). 

Using this as is, I get a deprecation notice triggered by Symfony:
```
 3x: Since symfony/framework-bundle 5.3: Accessing the "session" service directly from the container is deprecated, use dependency injection instead.
    2x in ClassesInstantiableTest::clearSessionAfter from ...
    1x in ClassesInstantiableTest::clearSessionBefore from ...
```
This notice has nothing to do with the tests I've written, the notices  are instead triggered by the way `Shopware\Core\Framework\Test\TestCaseBase\SessionTestBehaviour` (used by the ubiquitous`IntegrationTestBehaviour`) is written.

[The core `phpunit.xml.dist`](https://github.com/shopware/platform/blob/891e14d48ac916c649a1bb29ac87d75ef0feca61/phpunit.xml.dist#L75) supresses Symfony deprecations by setting 
```xml
<env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled" />
```
I propose that the recommended PHPUnit configuration files for plugins do the same, to avoid confusion.

(In the process of fixing this, the schema version used by the recommended configuration file should probably be adjusted, as well - the version in the current docs uses 7.1, while core uses 9.5.)

svlauer:
Closing this issue as the associated PR has been merged.

PR #502, opened 2022-03-02
500-ignore-symfony-deprecations-in-recommended-phpunit-config

svlauer:
[The core phpunit.xml.dist](https://github.com/shopware/platform/blob/891e14d48ac916c649a1bb29ac87d75ef0feca61/phpunit.xml.dist#L75) ignores Symfony deprecations.

This PR includes the directive for doing the same to the recommended `phpunit.xml.dist` on [the doc page on unit testing plugins](https://developer.shopware.com/docs/guides/plugins/plugins/testing/php-unit#phpunit-configuration).

Not doing this creates confusing deprecation warnings on running plugin tests (confusing because the warnings have nothing to do with the tests defined by the plugin), see [this issue](https://github.com/shopware/docs/issues/500).

While I was at it, I took the liberty to bump the schema version used by the recommended `phpunit.xml.dist` - since shopware requires phpunit 9.5, using the more current schema seems appropriate (and [this is also what core does](https://github.com/shopware/platform/blob/891e14d48ac916c649a1bb29ac87d75ef0feca61/phpunit.xml.dist#L4)).

### DIFF
--- a/guides/plugins/plugins/testing/php-unit.md
+++ b/guides/plugins/plugins/testing/php-unit.md
@@ -27,7 +27,7 @@ Here's an example configuration for the development template:
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/7.1/phpunit.xsd"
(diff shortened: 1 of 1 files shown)


shyim (approved):

elkmod:
Thanks for your contribution @svlauer

PR #507, opened 2022-03-04
fix: SyntaxError object instead of array

RafaelKr:


Christian-Rades:
Thank you for contributing!

Christian-Rades (approved):

PR #514, opened 2022-03-11
WIP: NEXT-18117 - Add app base url to manifest reference

fschmtt:


resources/references/app-reference/manifest-reference.md
```diff
@@ -59,6 +59,8 @@
     </webhooks>
     <!-- Optional, can be omitted if the administration should not be extended -->
     <admin>
+        <!-- Entry point for the Admin Extension API (since 6.x.x.x) -->
```
fschmtt, line 62: @keulinho I've noticed that the docs were still missing this new tag. We'll have to fill the `since <version>` once it's released 🙂

keulinho: that's why it is not documented as of now as you can not really use it now, we should add it to the docs when we document the whole admin extension api.

so i don't know if we should add it, as it is basically useless for now, otherwise i would add `6.4.8.0` as the since version, because the tag is supported from then on

fschmtt: Well, we could just leave this PR open IMO until we need it then. Of course, it doesn't make much sense to document a new tag, that has no effect yet 😅 

If there's a ticket to document the whole feature, feel free to close this PR.

PR #516, opened 2022-03-22
Fix Internal links for Custom SEO Url doc

osrecio:
Fix Internal links for Custom SEO Url documentation


elkmod:
Thanks for the fix @osrecio

PR #523, opened 2022-03-25
Use "tested on" instead of min requirements

kermie:
Presently, we only show min system requirements in our Documentation. Also in installation steps (when server is already running and configured), only minimum server requirements are checked.

Compatibilities with new PHP versions can only be found in change log when digging for it which is not very beginner friendly. To resolve that, the idea is to show versions that are tested in our CI. Additionally, we propose to report issues with newer versions in our issue tracker.

### DIFF
--- a/guides/installation/overview.md
+++ b/guides/installation/overview.md
@@ -54,7 +54,7 @@ Although Shopware 6 supports most UNIX like environments, we recommend using **U
 
 PHP
 
-* tested on 7.4.3, 8.0 and 8.1
+* Tested on 7.4.3, 8.0 and 8.1
 * `memory_limit` 512M minimum
 * `max_execution_time` 30 seconds minimum
 * Extensions:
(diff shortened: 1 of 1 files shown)


guides/installation/overview.md
```diff
@@ -54,7 +54,7 @@ Although Shopware 6 supports most UNIX like environments, we recommend using **U
 
 PHP
 
-* 7.4.3 or higher
+* tested on 7.4.3, 8.0 and 8.1
```
JoshuaBehrens, line 57: Any other line (that starts with non-technical text) starts with an upper-case character. Should it be normalized?

JoshuaBehrens (changes requested):

guides/installation/overview.md
Christian-Rades: I agree it should be uniform

Christian-Rades (changes requested):
Please begin Bullet points with uppercase letters

PR #534, opened 2022-04-01
App extension SDK starter guide

elkmod:


guides/plugins/apps/starter/starter-admin-extension.md
```diff
@@ -0,0 +1,149 @@
...
+
+```
+
+### Start the local development server
+
+Next, we need to start the live server so you don't always have to reload the page manually.
+
+```bash
+cd custom/apps/ListingExtension
+npm install -g live-server
```
shyim, line 86: You can directly use the live-server no need for ngrok.

elkmod: Is it also applicable for App development with cloud-stores or will we run into CORS issues then?

guides/plugins/apps/starter/starter-admin-extension.md
```diff
@@ -0,0 +1,155 @@
...
+mkdir src
+touch src/index.html
+```
+
+{% code title="index.html" %}
+```html
+<!doctype html>
+<html>
+    <head>
+        <script src="https://unpkg.com/@shopware-ag/admin-extension-sdk/cdn"></script>
```
shyim, line 84: Should we pin here to a specific version? So this doc does not break when we break the API?

guides/plugins/apps/starter/starter-admin-extension.md
```diff
@@ -0,0 +1,155 @@
...
+    </meta>
+    <admin>
+        <base-app-url>http://127.0.0.1:8080</base-app-url>
+    </admin>
+</manifest>
+```
+{% endcode %}
+
+Since the URL passed is only available locally, you will only be able to see changes on your own machine.
+If you want to share it, you need to host the entry point file somewhere or use services to expose local files as public URLs, such as [ngrok](https://ngrok.com/).
```
Christian-Rades, line 131: Maybe qualify this as a development solution. For example, a CDN or other static site hosting can distribute the prod version.

guides/plugins/apps/starter/starter-admin-extension.md
```diff
@@ -0,0 +1,155 @@
...
+
+First of all we need to create the app "wrapper", the so-called app manifest. It is just a single XML file with some basic configuration.
+
+### Create manifest file
+
+First of all, we create a new directory that contains our project.
+
+```bash
+mkdir ListingExtension
+```
```
Christian-Rades, line 31: I'm not a big fan of the bash commands. Maybe a file tree would be more visual. This also reduces cluttering

Christian-Rades (changes requested):

Christian-Rades (approved):

jleifeld (approved):

PR #536, opened 2022-04-03
Fix Add cms block Storefront representation guide

dpawelec:
The guide contains a wrong filename that does not match the previous steps. This changes fix it.


elkmod:
Thanks for your contribution @dpawelec

elkmod (approved):

PR #538, opened 2022-04-05
B2B-870 - B2B-Suite administration docs

dfrancos-hub:
This doc provides only an overview of the administration modules. When the Quick Start and the Offer guides are created, links to those pages will be also added here.

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -40,6 +40,7 @@
   * [Professional Edition](products/editions/professional-edition.md)
   * [Enterprise Edition](products/editions/enterprise-edition/README.md)
     * [B2B Suite](products/editions/enterprise-edition/b2b-suite.md)
+      * [Administration](products/editions/enterprise-edition/b2b-suite/administration/administration.md)
     * [Search](products/editions/enterprise-edition/search/README.md)
       * [Installation](products/editions/enterprise-edition/search/installation.md)
       * [Field Configuration](products/editions/enterprise-edition/search/field-config.md)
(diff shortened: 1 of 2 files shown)


bschulzebaek:
@dfrancos-hub maybe we should consider linking the administration page to the end user docs, instead of showing those two screenshots. The user docs can be found [here](https://docs.shopware.com/de/shopware-6-de/enterprise-erweiterungen/b2b-suite-administration).

dfrancos-hub:
thanks @bschulzebaek ,you are right I think thats better. Although the Offer management module is only described in the german version of the user documentation. What do you think? should we keep the second screenshot or removing it anyways?

bschulzebaek:
Maybe you could reach out to the responsible docs team then, so they can provide an english version as well 👍 
I would like to prevent having the same information in different places, where both would require some maintenance.

dfrancos-hub:
Agree @bschulzebaek 👍, ok ill reach them. Then, any further changes here?

bschulzebaek:
I would just remove the two paragraphs regarding "Extended Customer Details" and "Offers Module
". We will probably need to document some specific components or concepts of the b2b administration code (yet never really was an issue in support tickets, iirc), but right now, we want to focus on migrating the old documentation, not documenting everything, thats possibly missing right now.

dfrancos-hub:
Hi @bschulzebaek I totally agree, actually those paragraph are there only because it was suggested in the ticket :)

dfrancos-hub:
@bschulzebaek done ;)

bschulzebaek (approved):

PR #540, opened 2022-04-07
Add performance tweaks page

shyim:


guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+
+We recommend setting up at least four Redis servers for the following resources:
+- Session + cart - [Read more](../performance/session.md)
+- cache.object - [Read more](../performance/caches.md#example-replace-some-cache-with-redis)
+- Lock + Increment storage - [Read more](../performance/increment.md)
+- Enqueue - [Read more](../infrastructure/message-queue.md#transport-redis-example)
+
+Instead of setting up a Redis server for `enqueue`, you can also work directly with [RabbitMQ](../infrastructure/message-queue.md#transport-rabbitmq-example)
+
+The PHP Redis extension should be preferred to be installed on the machine as of Predis to use a persistent Redis connection. Persistent connection opens per FPM process a Redis connection and uses it for all requests, which can help in high load scenarios as each request doesn't have to open and close a connection and may hit the system max open sockets
```
Christian-Rades, line 19:
```suggestion
The PHP Redis extension provides persistent Redis connections. Persistent connections can help in high load scenarios as each request doesn't have to open and close connections. Using non-persistent Redis connections can also hit the system's maximum of open sockets. Because of these limitations, the Redis extension is preferred over Predis.
```

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+monolog:
+    handlers:
+        main:
+            level: error  
+            buffer_size: 30
+        business_event_handler_buffer:
+            level: error
+```
+
+The `business_event_handler_buffer` handler logs flows. Setting it to `error` will disable the logging of flow activities that succeed
```
Christian-Rades, line 37:
```suggestion
The `business_event_handler_buffer` handler logs flows. Setting it to `error` will disable the logging of flow activities that succeed.
```

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Filesystem
+
+In a multi-app-server system, manage specific directories over a shared filesystem. This includes assets as well as theme files and private and public filesystems. The recommendation is an S3 compatible bucket instead of NFS shares or GCP backend.
```
Christian-Rades, line 41: Do we prefer S3 over GCP? Or Do we prefer S3 and GCP over NFS? This is unclear from this sentence

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+
+When you have a lot of app servers, you should consider using a reverse proxy cache like Varnish. Shopware offers default configuration for Varnish out of the box.
+
+[Read more](../infrastructure/reverse-http-cache.md)
+
+### logged-in / cart-filled
+
+By default, Shopware can no longer deliver complete pages from a cache for a logged-in customer or if products are in the shopping cart. As soon as this happens, the user sessions differ, and the context rules could be different depending on the user. The result is different content for each customer; a good example is the [Dynamic Access](https://store.shopware.com/swag140583558965t/dynamic-access.html) plugin.
+
+However, if the project does not require such functionality, pages can also be cached by the HTTP cache/reverse proxy. The HTTP cache significantly reduces the load for the system. To disable Cache invalidation in these cases:
```
Christian-Rades, line 61:
```suggestion
However, if the project does not require such functionality, pages can also be cached by the HTTP cache/reverse proxy. To disable Cache invalidation in these cases:
```

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+```yaml
+shopware:
+    cache:
+        invalidation:
+            http_cache: []
+```
+
+### Delayed invalidation
+
+For systems with a high update frequency for the inventory (products, categories), it is recommended to activate a delay for the cache invalidation. Suppose the instruction to delete the cache entries for a specific product or category occurs in the code. In that case, the cache is not deleted directly but processed by a background task afterward. However, if another process also tries to invalidate one of the cache entries of the first process, there will be no overlap, but the timer for the invalidation of this cache entry will only reset.
```
Christian-Rades, line 72:
```suggestion
A delay for the cache invalidation can be activated for systems with a high update frequency for the inventory (products, categories). Once the instruction to delete the cache entries for a specific product or category occurs, they are not deleted instantly but processed by a background task afterwards. Thus, if two processes invalidate the cache in quick succession, the timer for the invalidation of this cache entry will only reset.
```

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+    mail:
+        update_mail_variables_on_send: false
+```
+
+[Read more](https://github.com/shopware/platform/blob/trunk/adr/2022-03-25-prevent-mail-updates.md)
+
+## Increment storage
+
+The increment storage is used to store the state and display it in the Administration.
+This storage increments or decrements a given key in a transaction-safe way, which causes locks upon the storage. Therefore, we recommend moving this load to a separate Redis:
```
Christian-Rades, line 124:
```suggestion
This storage increments or decrements a given key in a transaction-safe way, which causes locks upon the storage. Therefore, we recommend moving this source of server load to a separate Redis:
```

guides/hosting/performance/performance-tweaks.md
```diff
@@ -0,0 +1,208 @@
...
+```yaml
+framework:
+    lock: 'redis://host:port'
+```
+
+[Read more](../performance/lock-store.md)
+
+## Sending mails with the Queue
+
+Shopware sends the mails by default synchronously. This process can take a while on load or when the remote SMTP server is struggling. For this purpose, it is possible to handle the mails in the message queue. To enable this, add the following config to your config
```
Christian-Rades, line 157:
```suggestion
Shopware sends the mails by default synchronously. This process can take a while when the remote SMTP server is struggling. For this purpose, it is possible to handle the mails in the message queue. To enable this, add the following config to your config
```

Christian-Rades (changes requested):
Also, we should rename `enterprise setup` to `high-load setup`,  `large scale setup` or something of that kind.

Christian-Rades (approved):

guides/hosting/installation-updates/high-load-setup.md
```diff
@@ -0,0 +1,76 @@
+# Enterprise project setup
```
JoshuaBehrens, line 1: The page title is still enterprise project setup. Will this work good together with the search?

JoshuaBehrens: @Christian-Rades

PR #545, opened 2022-04-13
Update add-cms-block.md for more consistency

marcelbrode:


marcelbrode:
@elkmod Something missing here?

PR #563, opened 2022-04-26
WIP: PaaS product page

elkmod:


products/paas/fastly.md
```diff
@@ -0,0 +1,7 @@
+# Fastly
+
+
+1. Make sure you have at least Shopware 6.4.11.0
+2. Make sure `FASTLY_API_TOKEN` and `FASTLY_SERVICE_ID` are set in the environment or contact Platform.sh when its missing.
```
shyim, line 5: maybe contact the Support to get rid of it 🤔

products/paas/fastly.md
```diff
@@ -0,0 +1,7 @@
+# Fastly
+
+
+1. Make sure you have at least Shopware 6.4.11.0
```
shyim, line 4: We should explain what the benefits are of using Fastly 🤔

products/paas/repository.md
```diff
@@ -0,0 +1,82 @@
+# Repository
+
+The source code of your project will reside in a git-based VCS repository. You can start with a plain project, however we suggest starting with a fork of the [shopware/paas](https://github.com/shopware/paas) setup template from GitHub. You will find out more about the setup template in the [Setup Template](setup-template.md) section.
+
+## Fork and clone setup template
```
shyim, line 5: There are also integrations like for GitLab. I guess that is very important for agencies

products/paas/setup-template.md
```diff
@@ -0,0 +1,107 @@
...
+
+This section contains configuration for environment variables or server settings. General store settings and configuration are set here. In this place you can inject custom environment variables or enable feature flags.
+
+Variables in the `env` section are automatically injected as environment variables.
+
+### hooks
+
+Lifecycle hooks are custom scripts that are called during your build and deploy processes. See more on the [deployment process](./build-deploy#push-main-branch).
+
+#### build hook
```
shyim, line 49: Very important to say:

- In this step you don't have access to all services, only the file system is available
- All files which are not in a `mount` are here only writeable and later steps readonly
- The best practise is to do the most stuff in the `build` as the `deploy` is blocking the web traffic

products/paas/setup-template.md
```diff
@@ -0,0 +1,107 @@
...
+
+This script is called during the build process and builds your applications assets and disables the UI installer. You can customize this script if you need.
+
+#### deploy hook
```
shyim, line 53: Very important that is step is fast as possible. In the execution duration, no traffic is handled

shyim, line 53: Maybe we should also mention `post_deploy`. 
This hook is executed after the router is available and serves requests

products/paas/fastly.md
elkmod: what are the benefits?

shyim (approved):

PR #565, opened 2022-04-27
Wip: Guide for basic reading and Writing of data in Apps

Christian-Rades:


guides/plugins/apps/starter/product-translator.md
```diff
@@ -0,0 +1,363 @@
...
+
+{% code title="release/manifest.xml" %}
+``` xml
+<?xml version="1.0" encoding="UTF-8" ?>
+<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/platform/master/src/Core/Framework/App/Manifest/Schema/manifest-1.0.xsd">
+    <meta>
+    <!-- ... -->
+    </meta>
+    <setup>
+        <registrationUrl>http://127.0.0.1:8000/register</registrationUrl>
```
elkmod, line 110: Mention usage of ngrok URL here - Cloud development should be assumed as the "standard" case in this guide. Exceptions are always formulated as "When you're using a self-managed ... "

guides/plugins/apps/starter/product-translator.md
```diff
@@ -0,0 +1,363 @@
...
+        );
+    }
+```
+{% endcode %}
+
+Note that the hash of the original description gets saved as a value in the 
+custom fields of the product entity. This is possible without any further config
+due to the fact that all custom fields are schema-less.
+
+The implementation of the `translate` method is disregarded in this example, you might perform an additional lookup through a translation API service to implement it.
```
elkmod, line 363: Next steps?

elkmod (changes requested):

guides/plugins/apps/starter/product-translator.md
```diff
@@ -0,0 +1,363 @@
...
+        <webhook name="deleted" url="http://127.0.0.1:8000/registration/remove" event="app.deleted"></webhook>
+        <webhook name="product-update" url="http://127.0.0.1:8000/product/written" event="product.written"></webhook>
+    </webhooks>
+</manifest>
+```
+{% endcode %}
+
+These two webhooks provide a way for stores to notify your app server about events that occured.
+
+The `deleted` webhook is already provided by the `src/Controller/RegistrationController.php` controller in the app template. I notifies the server that a store has deleted the app.
```
elkmod, line 170:
```suggestion
The `deleted` webhook is already provided by the `src/Controller/RegistrationController.php` controller in the app template. It notifies the server that a store has deleted the app.
```

guides/plugins/apps/starter/product-translator.md
```diff
@@ -0,0 +1,398 @@
...
+
+The setup is as simple as calling the following command (after installing ngrok)
+
+```sh
+ngrok http 8000
+```
+
+This will expose your Symfony server on a public URL, so the cloud store can communicate with your app.
+
+## Creating the manifest
```
elkmod, line 69: Please improve readability by adding 3rd-level sub-headings where applicable.

guides/plugins/apps/starter/product-translator.md
```diff
@@ -0,0 +1,398 @@
...
+{% endcode %}
+
+These two webhooks provide a way for shops to notify your app server about events that occurred.
+The `src/Controller/RegistrationController.php`  controller in the app template already provides the `deleted` webhook. It notifies the server that a shop has deleted the app.
+
+The `product-update` webhook is the path on which your app server will be notified about any product updates in the stores, like changing the description.
+
+This webhook needs a custom controller, which will be the next part of this guide.
+
+## Implementing the translation
```
elkmod, line 171: Please improve readability by adding 3rd-level sub-headings where applicable

elkmod: Also I'm thinking "Implementing the translation" is not the correct heading, as the actual translation is not a part of the guide. Maybe more like "Implementing the webhook listener" (not happy with that one either, but it's closer to what we're describing).

elkmod (changes requested):

elkmod (approved):

PR #575, opened 2022-05-10
NEXT-20790 - documentation for scss declaration via config xml

ssltg:
Documentation for a new way to declare plugin config variables as SCSS variables.

### DIFF
--- a/guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
+++ b/guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
@@ -1,0 +1,238 @@
+# Add SCSS variables
+
+## Overview
+
+In order to add SCSS variables to your plugin, you can configure fields in your `config.xml` to be exposed as scss variables.
+
+We recommend to use the declaration of SCSS variables via the `config.xml` as described [here](./add-scss-variables.md) but you can still use a subscriber if you need to be more flexible as described below.
+
+## Prerequisites
+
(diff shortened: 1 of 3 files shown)


guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
```diff
@@ -0,0 +1,238 @@
+# Add SCSS variables
+
```
elkmod, line 2: If this only applies from a specific SW version, please add the version hint here.

guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
```diff
@@ -0,0 +1,238 @@
+# Add SCSS variables
```
elkmod, line 1: Please add this file to the global `SUMMARY.md` - otherwise it won't show up in the navigation.

elkmod (changes requested):
Missing `SUMMARY.md` entry and minor style changes.

guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
ssltg: This already applies to older versions, it was just moved because it is no longer the standard.

guides/plugins/plugins/storefront/add-scss-variables-via-subscriber.md
ssltg: added

elkmod (approved):

PR #576, opened 2022-05-11
Update systemd service template

saibotd:
A missing `WorkingDirectory` led to errors with the export:  
`The stream or file "//var/log/app.log" could not be opened in append mode: failed to open stream: Permission denied`  
Reason for this is probably path traversing issues when running directly from `/var/www/html/bin` or missing `.env` variables.  
`RestartSec=always` is just plain wrong here - should be a numerical or omitted completely.


Christian-Rades (approved):

Christian-Rades:
Thanks for contributing :tada:

PR #584, opened 2022-05-18
NEXT-21008 - Add prepared and refund to payment app guide

mstegmeyer:


guides/plugins/apps/payment.md
```diff
@@ -204,6 +218,107 @@ $response = [
...
+
+For this, you have two calls available during the order placement, the `validate` call to verify, that the payment reference is valid and if not, stop the placement of the order, and the `capture` call, which then allows the payment to be processed to completion after the order has been placed and persisted.
+
+Let's first talk about the `validate` call. Here, you will receive three items to validate your payment. The `cart` with all its line items, the `requestData` from the `CardOrderRoute` request and the current `salesChannelContext`. This allows you to validate, if the payment reference you may have given your payment handler via the Storefront implementation is valid and will be able to be used to pay the order which is about to be placed. The array data you may send as the `preOrderPayment` object in your response will be forwarded to your `capture` call, so you don't have to worry about identifing the order by looking at the cart from the `validate` call. If the payment is invalid, either return a response with an error response code or provide a `message` in your response.
```
lernhart, line 227:
```suggestion
Let's first talk about the `validate` call. Here, you will receive three items to validate your payment. The `cart` with all its line items, the `requestData` from the `CartOrderRoute` request and the current `salesChannelContext`. This allows you to validate, if the payment reference you may have given your payment handler via the Storefront implementation is valid and will be able to be used to pay the order which is about to be placed. The array data you may send as the `preOrderPayment` object in your response will be forwarded to your `capture` call, so you don't have to worry about identifying the order by looking at the cart from the `validate` call. If the payment is invalid, either return a response with an error response code or provide a `message` in your response.
```

guides/plugins/apps/payment.md
```diff
@@ -204,6 +218,107 @@ $response = [
...
+}
+```
+
+{% hint style="warning" %}
+Keep in mind that if the integration into the checkout process does not work as expected, your customer might not be able to use the prepared payment. This is especially valid for after order payments, since there the order already exists. For these cases, you should still offer a traditional synchronous / asynchronous payment flow. Don't worry, if you have set the transaction state in your capture call to anything but open, the asynchronous payment process will not be started immediately after the prepared payment flow.
+{% endhint %}
+
+## Refund
+
+With Shopware 6.4.12.0, we have also added basic functionality to be able to refund payments. Your app will need register captured amounts into Shopware for Shopware to be able to process a refund of a capture.
```
lernhart, line 291:
```suggestion
With Shopware 6.4.12.0, we have also added basic functionality to be able to refund payments. Your app will need to register captured amounts and create and persist a refund beforehand for Shopware to be able to process a refund of a capture.
```

guides/plugins/apps/payment.md
```diff
@@ -204,6 +218,107 @@ $response = [
...
+
+Similar to the other requests, you will receive on your `refund` call the data required to process your refund. This is the `order` with all its details and also the `refund` which holds the information on the `amount`, the referenced `capture` and, if provided, a `reason` and specific `positions` which items are being refunded.
```
lernhart, line 293:
```suggestion
Similar to the other requests, on your `refund` call, you will receive the data required to process your refund. This is the `order` with all its details and also the `refund` which holds the information on the `amount`, the referenced `capture` and, if provided, a `reason` and specific `positions` which items are being refunded.
```

lernhart (approved):

PR #585, opened 2022-05-19
B2B-868 - update b2b-suite example plugins

dfrancos-hub:


dfrancos-hub:
this is only the update of the plugins, the structure and links from docs will be done in another ticket: b2b-916

SimonVorgers (approved):

PR #589, opened 2022-05-23
Fix input field in tutorial

levin192:
Following the tutorial, the sw-text-field will show [Object object] instead of the value of the config field. Added the missing property here.


elkmod (approved):

elkmod:
Thanks for your contribution @levin192!

PR #592, opened 2022-05-26
Now the apps are allowed to create new entities

osrecio:
Based on the documentation Apps now are capables for create new entities: https://developer.shopware.com/docs/guides/plugins/apps/custom-data/custom-entities


elkmod:
Thank you for the contribution @osrecio !

elkmod (approved):

PR #593, opened 2022-05-30
Add database cluster page

shyim:


guides/hosting/infrastructure/database-cluster.md
```diff
@@ -0,0 +1,41 @@
+# Database Cluster
+
+To scale Shopware beyond the horizon, we recommend you to use a database cluster. A database cluster consists of multiple read-only servers that are managed by a single primeray instance. Shopware comes in default with an read write splitting of SQL queries. When an Insert/Update/Delete query is executed, the query is distributed to the primary server and the connection uses only the primary until the end of the connection, this will be also triggered by the ``executeStatement`` in the [DebugStack decoration](https://github.com/shopware/platform/blob/v6.4.11.1/src/Core/Profiling/Doctrine/DebugStack.php#L48).
+This makes it possible to write an record and read it directly, without taking care of the cluster in the Application.
```
aragon999, line 4:
```suggestion
This makes it possible to write or read a record, without taking the possibility into account that the read only child nodes might not be in sync with the primary node.
```

guides/hosting/infrastructure/database-cluster.md
```diff
@@ -0,0 +1,41 @@
+# Database Cluster
+
+To scale Shopware beyond the horizon, we recommend you to use a database cluster. A database cluster consists of multiple read-only servers that are managed by a single primeray instance. Shopware comes in default with an read write splitting of SQL queries. When an Insert/Update/Delete query is executed, the query is distributed to the primary server and the connection uses only the primary until the end of the connection, this will be also triggered by the ``executeStatement`` in the [DebugStack decoration](https://github.com/shopware/platform/blob/v6.4.11.1/src/Core/Profiling/Doctrine/DebugStack.php#L48).
```
aragon999, line 3:
```suggestion
To scale Shopware beyond the horizon, we recommend to use a database cluster. A database cluster consists of multiple read-only servers that are managed by a single primary instance. Shopware comes in default with an read write splitting of SQL queries. When an [`INSERT`/`UPDATE`/`DELETE`/...](https://github.com/shopware/platform/blob/v6.4.11.1/src/Core/Profiling/Doctrine/DebugStack.php#L48) query is executed, the query is distributed to the primary server and the connection uses only the primary node until the end of the connection. This will be ensured by the ``executeStatement`` in the [DebugStack decoration](https://github.com/shopware/platform/blob/v6.4.11.1/src/Core/Profiling/Doctrine/DebugStack.php#L48).
```

JoshuaBehrens, line 3: Is "beyond the horizon" the right choice of words? I mean Udo Lindenberg is not wrong with that and you want to express to overcome a limitation, but I would still think about it. Maybe a "even further" is a better one. Maybe you want to express instead that there are a lot of performance tweaks you can do and one of them is to upgrade your single instance to a clustered database.

JoshuaBehrens, line 3: > ``executeStatement``

Double backtick to single backtick?

JoshuaBehrens (changes requested):

guides/hosting/infrastructure/database-cluster.md
shyim: When i am in the flow i just write anything at once forgot to look later again into 😂

guides/hosting/infrastructure/database-cluster.md
```diff
@@ -0,0 +1,41 @@
...
+This makes it possible to write or read a record, without taking the possibility into account that the read-only child nodes might not be in sync with the primary node.
+
+## Preparing Shopware
+
+To make the Splitting most effective, we recommend to configure the following steps:
+
+### Using the optimal MySQL configuration
+
+Shopware does set by default some specific MySQL configuration to make sure that the database is optimized for the Shopware usage. 
+This variables are set in the cluster mode only on the read-only server. To make sure that Shopware works flawlessly these configuration must be configured directly on the MySQL server so these variables are set on any server.
```
elkmod, line 13:
```suggestion
These variables are set in the cluster mode only on the read-only server. To make sure that Shopware works flawlessly these configuration must be configured directly on the MySQL server so these variables are set on any server.
```

guides/hosting/infrastructure/database-cluster.md
```diff
@@ -0,0 +1,41 @@
...
+
+The following options should be set:
+
+- Make sure that `group_concat_max_len` is by default higher or equal to `320000`
+- Make sure that `sql_mode` doesn't contain `ONLY_FULL_GROUP_BY`
+
+After this change, you can set also `SQL_SET_DEFAULT_SESSION_VARIABLES=0` in the `.env` file so Shopware does not check for those variables in runtime.
```
elkmod, line 20:
```suggestion
After this change, you can set also `SQL_SET_DEFAULT_SESSION_VARIABLES=0` in the `.env` file so Shopware does not check for those variables at runtime.
```

guides/hosting/infrastructure/database-cluster.md
```diff
@@ -0,0 +1,41 @@
...
+As we learned in the beginning, Shopware uses read-only MySQL server until the first write attempt. To maximaze this behaviour it is hightly recommanded to outsource much as you can the writes on the Database. One of the easiest solutions is to use the Redis as storage for the Cart.
+To use Redis add following snippet to `config/packages/cart.yml`
+
+```yml
+shopware:
+    cart:
+        redis_url: 'redis://localhost:6379/0?persistent=1'
+```
+
+It is recommanded to use a persistent Redis connection to avoid connection issues in high load scenarios. There is also an `cart:migrate` command to Migrate the cart between MySQL and Redis so the users can continue to use the existing carts.
```
elkmod, line 33:
```suggestion
It is recommended to use a persistent Redis connection to avoid connection issues in high load scenarios. There is also a `cart:migrate` command to migrate the existing carts between MySQL and Redis so the migration does not influence end user experience.
```

elkmod (changes requested):

elkmod (approved):

PR #602, opened 2022-06-11
Fix number range docs

Isengo1989:
Typo in number_range -> see https://github.com/shopware/platform/blob/ba2f87d5413ffac3ee0489e3dcf999a359d7b3a3/src/Core/Framework/Resources/config/packages/shopware.yaml#L173


JoshuaBehrens:
@elkmod 
image

PR #609, opened 2022-06-15
DX-104 - First try to integrate ADRs

Christian-Rades:


JoshuaBehrens:
Are they just copied over? Although they are unlikely to change, they do change. Do we have an update strategy?

Christian-Rades:
@JoshuaBehrens Yes, the plan is to sync them through a Github action. But first I want to get the scripts right to format the files

shyim (approved):

PR #614, opened 2022-06-24
fixed typo

Donderda:
I think the correct syntax is `{% parent %}`, not `\{% parent %\}`


elkmod (approved):

elkmod:
Thanks for your contribution @Donderda

PR #641, opened 2022-07-21
Scott anderson programming examples and readme fixes

ScottA38:
@elkmod These are a couple of things I found in the docs that I think could be amended
For the phpunit config this was true for phpunit v9.5.2

Both are code changes, so a developers' oversight that the admendments are actually correct would be helpful - @marcelbrode ?


shyim (approved):

elkmod (approved):
Thanks for expanding @ScottA38 !

PR #644, opened 2022-07-26
NEXT-22576 - Add soft purge docs

shyim:


guides/hosting/infrastructure/reverse-http-cache.md
```diff
@@ -242,7 +246,215 @@ sub vcl_deliver {
 }
 ```
 
-To verify if it works, you can look for a new response header `X-Cache` in the http response. It shows you if it was a cache hit or miss.
+To verify if it works, you can look for a new response header `X-Cache` in the HTTP response. It shows you if it was a cache hit or miss.
+
+#### Soft purge
```
keulinho, line 251: does it work out of the box or is the latest version needed? :thinking:

shyim: for varnish it works out of the box with that config 😅 , you have to specify it in the varnish.vcl instead of shopware

keulinho (approved):

Christian-Rades (approved):

PR #657, opened 2022-08-11
Fix App Icon path

wirduzen-felix:
The Folder structure right above the manifest.xml shows, that the icon is actually called app-icon.png and not app.png. This might be confusing.
image


Christian-Rades (approved):

Christian-Rades:
Good Point, thank you!

PR #660, opened 2022-08-16
Update plugin-base-guide.md

### LINKED ISSUE #518: plugin:create nowhere to be seen in the "Plugin Base Guide"
[The Plugin Base Guide](https://developer.shopware.com/docs/guides/plugins/plugins/plugin-base-guide) explains how to manually create a plugin from scratch without relying on CLI's `create` command, seemingly suggesting by omission that there is no command for generating the plugin's skeleton. 
For experienced readers, that may be obvious. But assuming the target audience is wider, explicitly mentioning `plugin:create` within the "_Create the plugin_" subsection would do no harm. Not least because the presence of blog posts like [this one](https://www.matheusgontijo.com/2022/01/24/how-to-create-a-plugin-in-shopware-6) or [this other one](https://www.magespark.com/blog/post/how-create-plugin-shopware-6) on the web about precisely the same subject clearly points to that omission being a significant gap in the official dev docs.
In the worst case scenario, if the reader happens to be aware of other commands such as `theme:create`, the assumption may be that you can only scaffold plugins automatically as initial themes, with the subsequent unnecessary bloating the theme-specific code introduces for something not meant to be a theme in the first place. This was my case.

zaifastafa:
Fixes #518

### DIFF
--- a/guides/plugins/plugins/plugin-base-guide.md
+++ b/guides/plugins/plugins/plugin-base-guide.md
@@ -32,7 +32,7 @@ Now that you've found your name, it's time to actually create your plugin.
 
 Shopware provides a handy command that you can use to generate the plugin structure. Go to your shopware project's root directory and run the following command:  
 
-```
+```bash
 bin/console plugin:create SwagBasicExample
 ```
 
--- a/guides/plugins/plugins/plugin-base-guide.md
+++ b/guides/plugins/plugins/plugin-base-guide.md
@@ -30,6 +30,16 @@ Using a prefix for your plugin name is not just a convention we'd recommend, but
 
 Now that you've found your name, it's time to actually create your plugin.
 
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/plugin-base-guide.md
```diff
@@ -30,6 +30,16 @@ Using a prefix for your plugin name is not just a convention we'd recommend, but
 
 Now that you've found your name, it's time to actually create your plugin.
 
+Shopware provides a handy command that you can use to generate the plugin structure. Go to your shopware project's root directory and run the following command:  
+
+```
```
Christian-Rades, line 35:
```suggestion
```bash
```

Christian-Rades (changes requested):
Thank you! We require our code blocks to specify a language. Once that is done, we can merge your PR.

Christian-Rades (approved):

PR #663, opened 2022-08-22
NEXT-22890 - Additional the documentation for Add Action in Flow Builder

nguyenky:
Additional the docs for group action in Flow Builder


guides/plugins/plugins/framework/flow/add-flow-builder-action.md
```diff
@@ -225,30 +225,54 @@ export const ACTION = Object.freeze({
...
+
 export default {
-    ACTION,
+    ACTION, GROUP
 };
 ```
 
 {% endcode %}
 
+The default group is `general`, that mean if you define `GROUP` is `general` or don't define `GROUP`, your action will be into the `General` group in the list actions.
```
elkmod, line 237:
```suggestion
**Grouping Actions**

Using the `GROUP` constant, you can define a group that your action will show up in. The default group is `general`.
```

guides/plugins/plugins/framework/flow/add-flow-builder-action.md
```diff
@@ -225,30 +225,54 @@ export const ACTION = Object.freeze({
...
 ```jsx
-import { ACTION } from '../../constant/swag-example-plugin.constant';
+import { ACTION, GROUP } from '../../constant/swag-example-plugin.constant';
 
 const { Component } = Shopware;
 
 Component.override('sw-flow-sequence-action', {
+    computed: {
+        // Not necessary if you use an existing group
+        // Push the `groups` method in computed if you are define a new group
```
elkmod, line 260:
```suggestion
        // Push the `groups` method in computed if you are defining a new group
```

elkmod (approved):

PR #670, opened 2022-08-24
NEXT-21532 - Add docs for custom entity ce shorthand

keulinho:
https://gitlab.shopware.com/shopware/6/product/platform/-/merge_requests/8760

### DIFF
--- a/guides/plugins/apps/custom-data/custom-entities.md
+++ b/guides/plugins/apps/custom-data/custom-entities.md
@@ -58,3 +58,31 @@ you need the appropriate [permissions](../../../../resources/references/app-refe
     </permissions>
 </manifest>
 ```
+
+## Shorthand prefix
+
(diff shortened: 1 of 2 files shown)


guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -58,3 +58,27 @@ you need the appropriate [permissions](../../../../resources/references/app-refe
     </permissions>
 </manifest>
 ```
+
+## Shorthand prefix
+
+Since v6.4.15.0 it is possible to also use the `ce_` shorthand prefix for your custom entities to prevent problems with length restrictions of names inside the DB.
```
shyim, line 64: Maybe hint about that changing the table name deletes all data? 🤔

shyim (approved):

PR #683, opened 2022-09-21
fix: add missing form csrf handler

shyim:
JS plugin is missing for ajax mode


sushmangupta (approved):
Thanks for the contribution.

PR #702, opened 2022-09-28
Add admin app snippet documentation

marcelbrode:


MoritzKrafeld (approved):

guides/plugins/plugins/administration/adding-snippets.md
```diff
@@ -8,6 +8,10 @@ By default Shopware 6 uses the [Vue I18n](https://kazupon.github.io/vue-i18n/sta
 
 Normally you want to use snippets in you custom module. To keep things organized, create a new directory named `snippet` inside your module directory `<plugin root>/src/Resources/app/administration/src/module/<your-module>/snippet`. For each language you want to support, you need a JSON file inside here, e.g. `de-DE.json` and of course `en-GB.json`.
 
+{% hint style="info" %}
+Providing snippets for apps works the same as in plugins, with the difference of a more simplistic file structure. Also, please keep in mind that instead of plugins, App-Snippets are **not allowed** to override existing snippet keys. So please use the following path for vendor-prefixed app snippet files instead: `<app root>/Resources/app/administration/snippet`
+{% endhint %}
+
```
sushmangupta, line 14:
```suggestion
Normally you use snippets in your custom module. To keep things organized, create a new directory named `snippet` inside module directory `<plugin root>/src/Resources/app/administration/src/module/<your-module>/snippet`. For each language you want to support, you need a JSON file inside it, e.g., `de-DE.json`, `en-GB.json`.

{% hint style="info" %}
Providing snippets for apps works the same as in plugins but it has a more simplistic file structure. Also, unlike plugins, App-Snippets don't override existing snippet keys. So, use the following path for vendor-prefixed app snippet files: `<app root>/Resources/app/administration/snippet`
{% endhint %}
```

sushmangupta (changes requested):

guides/plugins/plugins/administration/adding-snippets.md
marcelbrode: I'm not 100% sure about your change about "App-Snippets are **not allowed** to override existing snippet keys", to "don't override snippet keys". Since I think it's important to mention that it's not allowed and will cause shopware to get (intended) errors in that case.  So I'd stay with "are not allowed" or maybe "cannot", if thats okay

I patched everything here, besides that point. Is this okay for you?

sushmangupta: Yes I am fine with it. Please use "**are not allowed**" itself and keep the rest of the other changes as suggested.

marcelbrode: Done! Thanks!

sushmangupta (approved):

elkmod (approved):

Issue #674, opened 2022-09-02
Regarding mistake on "Add CMS block" document of shopware 6


utsav1-cyber:
Hi,

On the "Add CMS block" documentation of shopware 6. There is a minor mistake in the documentation. The name of the scss file on the component folder is incorrect. 

There is a line "create it now by adding the file sw-cms-block-image-text-reversed.scss to your component directory."

"my" word is missing on the scss file.

The correct file name should be "sw-cms-my-block-image-text-reversed.scss".
![Screenshot_4](https://user-images.githubusercontent.com/77831507/188062445-cdab678b-2ea0-4ff4-a05a-faa0f400acae.png)

https://developer.shopware.com/docs/guides/plugins/plugins/content/cms/add-cms-block

PR #704, opened 2022-09-30
fixed typo in cms-block guide

NinjaArmy:
There is a word missing in the docs. 
See: [This issue](https://github.com/shopware/docs/issues/674 )


NinjaArmy:
Hacktoberfest label would be appreciated

tobiasberge (approved):
Hi, looks fine from my side. 👍 

First I was a bit confused because in the diff of this PR the other areas are without the "my". But then I realized there was another [PR](https://github.com/shopware/docs/pull/545/files) a while ago which already fixed a bunch of those mistakes and it is also with the "my" in the live docs. Probably this was just not rebased or something, not sure why the diff shows old stuff.

PR #705, opened 2022-10-02
NTR - Refactor add-translations.md to also consider apps

marcelbrode:


guides/plugins/plugins/storefront/add-translations.md
```diff
@@ -6,25 +6,24 @@ In this guide you'll learn how to add translations to the storefront and how to
...
+Shopware 6 is able to load your snippet files automatically if you stick to a convention regarding file structure and naming. You have to store your snippet files in the `<extension root>/src/Resources/snippet` directory of your extension, but you can use further subdirectories if you want to. Use the naming pattern `<name>.<locale>`, where you can freely define the name part. The locale part must map to the ISO string of the supported locale in this snippet file. If you provide base translations, meaning that you ship translations for a whole new language, you indicate it with a `.base` suffix in your file name. Keep in mind that in this case you also have to use the name, so your complete filename may look like this: `<name>.<locale>.base.json`.
```
sushmangupta, line 13:
```suggestion
Shopware 6 automatically loads your snippet files when a standard file structure and a naming convention are followed. To do so, store your snippet files in the `<extension root>/src/Resources/snippet` directory of your extension. Also, you can further use subdirectories if you want to. Use `<name>.<locale>` as the naming pattern. The name can be freely defined, while the locale must map to the ISO string of the supported locale in this snippet file. If you provide base translations (ship translations for a whole new language), indicate it with a `.base` suffix in your file name. Now the filename convention to be followed looks like this `<name>.<locale>-base.json` - for example, *example.de-DE.json*.
```

In `<name>.<locale>-base.json`  I believe, the locale must be followed by a hyphen rather than a period. If it is not so, then please correct it and provide another example.

sushmangupta (changes requested):

guides/plugins/plugins/storefront/add-translations.md
marcelbrode: Nope. It definitly doesn't. I adjusted it, so you can have a look at it :)

marcelbrode: I also added an example for the first possible naming pattern and used "de-AT" instead of "de-DE" sind you won't ever provide a new language für germany-german.

guides/plugins/plugins/storefront/add-translations.md
```diff
@@ -6,25 +6,26 @@ In this guide you'll learn how to add translations to the storefront and how to
...
+Shopware 6 automatically loads your snippet files when a standard file structure and a naming convention are followed. To do so, store your snippet files in the `<extension root>/src/Resources/snippet` directory of your extension. Also, you can further use subdirectories if you want to. Use `<name>.<locale>` as the naming pattern. The name can be freely defined, while the locale must map to the ISO string of the supported locale in this snippet file - for example `example.de-DE.json`.
+
+If you provide base translations (ship translations for a whole new language), indicate it with a `.base` suffix in your file name. Now the filename convention to be followed looks like this `<name>.<locale>.base.json` - for example, `example.de-AT.base.json`.
```
sushmangupta, line 15:
```suggestion
In case you want to provide base translations (ship translations for a whole new language), indicate it with the suffix `.base` in your file name. Now the filename convention to be followed looks like this `<name>.<locale>.base.json` - for example, `example.de-AT.base.json`.
```

sushmangupta (approved):

PR #707, opened 2022-10-04
Fix typo error

shoutershub:
Typo Error leading to confusion on 

line 94 
Line 97 & 
Line 106

### DIFF
--- a/guides/plugins/plugins/framework/flow/add-flow-builder-action.md
+++ b/guides/plugins/plugins/framework/flow/add-flow-builder-action.md
@@ -91,10 +91,10 @@ class CreateTagAction extends FlowAction
 
     public function handle(FlowEvent $event): void
     {
-        // config is the config data when created a flow sequence
+        // config is the "Configuration data" you get after you create a flow sequence
         $config = $event->getConfig();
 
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/framework/flow/add-flow-builder-action.md
```diff
@@ -103,7 +103,7 @@ class CreateTagAction extends FlowAction
 
         $tags = $config['tags'];
 
-        // just a step to make sure you're dispatching correct action
+        // just a step to make sure you're dispatching the correct action
```
sushmangupta, line 106: Please be aware to not use contractions such as you're, we're, let's, it's. But you are free to use negative contractions such as don't, can't, etc... 
We are in the process to publish documentation guidelines to keep all the contributors aware of certain writing styles.

```suggestion
        // just a step to make sure you are dispatching the correct action
```

shoutershub: Duly noted, i will make corrections.

sushmangupta (approved):
Thanks for your contribution.

PR #713, opened 2022-10-10
NTR - Removed unused images

AegeliusFuneris:
I checked these images for usage. Since they did not appear anywhere, I deleted them.


marcelbrode (approved):
Have checked it. Those images can be removed safely :+1:

PR #715, opened 2022-10-10
Update composer installation to use 6.4

JoshuaBehrens:
Someone starting with the composer install guide should not install 6.3 by following the instructions


sushmangupta (approved):
Thanks for your contribution.

PR #718, opened 2022-10-11
Add locale to directory example

Schrank:


sushmangupta (approved):
Thanks for the update.

PR #722, opened 2022-10-12
NTR - added an example for pluralization using snippets

AegeliusFuneris:
especially for newbies a nice information   :)

### DIFF
--- a/guides/plugins/plugins/administration/adding-snippets.md
+++ b/guides/plugins/plugins/administration/adding-snippets.md
@@ -19,6 +19,7 @@ Each language then receives a nested object of translations, so let's have a loo
     "swag-example": {
         "nested": {
             "value": "example"
+            "examplePluralization": "1 Product | {n} Products"
         },
         "foo": "bar"
     }
@@ -59,7 +60,7 @@ The same `$tc` helper function can be used in the templates to access translatio
 
 {% raw %}
 
-```text
+```twig
 {% block my_custom_block %}
     <p>
        {{ $tc('swag-example.general.myCustomText') }}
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/administration/adding-snippets.md
```diff
@@ -69,6 +70,23 @@ The same `$tc` helper function can be used in the templates to access translatio
 
 {% endraw %}
 
+Another feature of `$tc` is pluralization. Using a `|` in snippets, you can provide translations dependending on a number. The first part would be shown as the singular expression, while the second takes care of your plural cases.
```
sushmangupta, line 73:
```suggestion
Another feature of `$tc` is pluralization. Use a `|` in the snippet to provide translations depending on the number. The first part shows singular expressions, while the second takes care of plural cases.
```

guides/plugins/plugins/administration/adding-snippets.md
```diff
@@ -69,6 +70,23 @@ The same `$tc` helper function can be used in the templates to access translatio
...
+{% block my_custom_block %}
+    <p>
+       {{ $tc('swag-example.nested.examplePluralization', products.length) }}
+    </p>
+{% endblock %}
+```
+
+{% endraw %}
+
+If you provide `1` as a parameter to `$tc()`, the text `One Product` would be rendered. Otherwise, if you provide e.g. `4` as a parameter you'll get `4 Products`.
```
sushmangupta, line 88:
```suggestion
If you provide `1` as the parameter to `$tc()`, the text `One Product` would be rendered. For any other value greater than 1, the number itself is shown — for example, `4 Products`.
```

sushmangupta (changes requested):

guides/plugins/plugins/administration/adding-snippets.md
marcelbrode: I guess "singular expression" is right, since he means the grammatical singular and there's just one singular expression right there

guides/plugins/plugins/administration/adding-snippets.md
AegeliusFuneris: Thank you for your suggestion. Since there are more than one parameters I specified it.

sushmangupta (approved):

PR #767, opened 2022-10-27
feat: remove mamp

shyim:
- Is a paid product so our users have to buy it
- Does not have `pcre.jit` for PHP by default so Shopware is slow
- They somehow compile PHP that it's soo slow

See also https://kleinmann.org/2022/06/17/php-development-environment-with-macos-homebrew-and-caddy/ many people have issues with that


JoshuaBehrens:
When this is missing, some might think you did not try mamp yet. In some parts of the docs there is a hint, that e.g. some node versions or mysql versions are not supported. Maybe you want to express that mamp is a bad idea for shopware

shyim:
@JoshuaBehrens so like adding mamp as incompatible to the overview?

JoshuaBehrens:
Yeah something like that. Maybe keep the mamp page but state right away that it is a bad idea. best case already in the menu structure. This way one does not think it will work and then click on the page to see that it won't

elkmod:
I agree with @JoshuaBehrens, that instead of removing we can rather indicate that we discourage this as an option. For someone who is still curious, they might look into it to find our they shouldn't use it.

Christian-Rades:
I would move the old guides into a common folder called `Legacy Setups` or something like this. So that people can still find the article while discouraging its use.

Christian-Rades (approved):

PR #770, opened 2022-11-04
feat(dal): added flag descriptions from source

AubreyHewes:
DAL Flags documentation was very poor.

I Simply copy pasted the flag descriptions from the source (and added `ApiAware` flag). This gives more insight in to the meaning of the flags than just a "this flag is this flag table".

Todo is probably a better description with examples per flag on how/why you would use them.


Christian-Rades (approved):

Christian-Rades:
Thank you for your contribution 🎉

PR #784, opened 2022-11-25
fix: condition decorator file name

saitho:
The name of the JS file for adding the condition to administration seems to be `condition-type-data-provider.decorator.js` as written in the code example.

I'm new to Shopware development. My rule showed up in administration after I changed the file name.
Which does not really make sense as I believe this to be actually dependent on how the file is included in main.js.

However Shopify Core rules also name the file `condition-type-data-provider.decorator.js`, so I guess docs should be consistent here.


elkmod:
Hey @saitho, we're having a look at it.  For context I'm adding the related documentation article - https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules#defining-restrictions

Maybe in the meantime you can have a look at how to add rule conditions with the [App system](https://developer.shopware.com/docs/guides/plugins/apps/rule-builder/add-custom-rule-conditions).

elkmod:
@saitho, we are not able to reproduce the error. The name of the decorator file is irrelevant, as long as the import statement references the correct file name. It is very likely that the issue you've experienced was related to something else - e.g. having a type error in the import statement and the admin file watcher freezing as a result.

If you have these types of issues, restarting the admin watch script can solve the problem. We're closing this PR, because it technically doesn't make any difference. Still, thank you for reporting and contributing.

PR #796, opened 2022-12-16
EntityRepositoryInterface is being deprecated

zetno:
Changed EntityRepositoryInterface to EntityRepository because EntityRepositoryInterface is being deprecated.


Christian-Rades:
Thank you for your contribution :+1: . But Please have some patience we will merge it once we have a 6.5 Branch officially.

PR #797, opened 2022-12-18
Updated resources section

sushmangupta:
Related ticket : [DX-209](https://shopware.atlassian.net/browse/DX-209)


elkmod (changes requested):
All files in `references/adr/*` are auto-generated, so the changes will be overridden with the next deployment.

elkmod (approved):

PR #818, opened 2023-01-10
Rename occurrences of "http" to "HTTP" in text passages

fschmtt:


concepts/framework/http_cache.md
```diff
@@ -52,20 +52,20 @@ Shopware uses several client cookies to differentiate the requests. This allows
 
 This cookie will be set when the non-logged-in customer with an empty cart changes the current currency. Why does Shopware need a separate cookie for currency? It allows us to maximize the cache hits for non-logged-in customers as we separate the cache as less as possible.
 
-### sw-cache-hash
+#### sw-cache-hash
```
sushmangupta, line 55:
```suggestion
### sw-cache-hash
```

concepts/framework/http_cache.md
```diff
@@ -52,20 +52,20 @@ Shopware uses several client cookies to differentiate the requests. This allows
...
 
 This cookie replaces the `sw-currency` cookie and contains the active rules and active currency. This cookie will be set when the active rules do not match the default anymore \(e.g., customer login/items in cart\).
 
-### sw-states
+#### sw-states
```
sushmangupta, line 59:
```suggestion
### sw-states
```

concepts/framework/http_cache.md
sushmangupta: As per documentation guidelines and linter formatter, heading levels can't be skipped.

concepts/framework/http_cache.md
sushmangupta: As per documentation guidelines and linter formatter, heading levels can't be skipped.

sushmangupta (changes requested):

concepts/framework/http_cache.md
fschmtt: Ah I see - so the goal is to not have them listed in the TOC on the right sidebar, right?

fschmtt: The previous `### Cache Cookies` is a `h3`, so I thought it would make sense to put each cookie explanation under it as a `h4` 🤔 

Am I missing something?

sushmangupta: Yeah, you are right, that info was actually collapsed. On expanding it, I see what you actually mean. We will go with your suggestion. Thank you.! :)

sushmangupta (approved):
Thanks for your contribution! :)

PR #823, opened 2023-01-18
NEXT-24864 - Add docs for Flow Sharing in Flow Builder

daothithientamm:


sushmangupta:
This content should be part of https://docs.shopware.com/en/shopware-6-en/features/share-flows as this is a user instruction for using the admin panel. Hence closing.

PR #841, opened 2023-02-01
DX-274 - Update image paths

bojanrajh:
This PR updates image (and .zip) URLs to the correct relative path and fixes some broken references to avoid issues with vitepress (https://github.com/shopware/docs/pull/824) and to correctly display images (in code editors, on GitHub, ...) - GitBook autocorrects incorrect paths.


resources/references/adr/checkout/2021-10-01-payment-flow.md
```diff
@@ -21,15 +21,15 @@ The synchronous payment is intended to execute a payment immediately after the o
 
 The following diagram shows a happy case sequence of a synchronous payment handling. The error handling is described [here](#after-order-payment-error-case) 
 
-![Synchronous Payment](../assets/payment-flow/synchronous-payment.png)
+![Synchronous Payment](../../../../.gitbook/assets/adr/payment-flow/synchronous-payment.png)
```
Christian-Rades, line 24: This is correct, but it should have been generated like this by https://github.com/shopware/docs/blob/master/.github/scripts/update-adrs.sh

resources/references/adr/performance/2022-03-25-profiler-integrations.md
```diff
@@ -107,7 +107,7 @@ class CartCalculator
 ```
 
 These spans are then displayed in the timeline of the corresponding profilers:
-![](../../../../.gitbook/assets/adr/tideways_benchmark.PNG "Tideways benchmark")
+![](../../../../.gitbook/assets/adr/tideways_benchmark.png "Tideways benchmark")
```
Christian-Rades, line 110: this should be lowercased automatically  in https://github.com/shopware/docs/blob/master/.github/scripts/update-adrs.sh

resources/references/adr/checkout/2021-10-01-payment-flow.md
bojanrajh: Technically, images have incorrect links in `shopware/platform` - [[2x here]](https://github.com/shopware/platform/blob/trunk/adr/checkout/2021-10-01-payment-flow.md), [[1x here]](https://github.com/shopware/platform/blob/trunk/adr/performance/2021-11-09-increment-pattern.md). Then I can cover both cases (`../assets/` and `./assets/`) correctly in the `update-adrs.php`.

resources/references/adr/performance/2022-03-25-profiler-integrations.md
bojanrajh: With previous comment in mind, since we'll need to create a PR in `shopware/platform`, I can as well fix this image there and then update ADRs.

Christian-Rades: Good Idea. I hadn't realised the source adrs were faulty.

elkmod (approved):

PR #869, opened 2023-02-21
Change variable naming

jeboehm:
ManyToOneAssociationField gives only one of each type, this is why I changed the variable naming


sushmangupta (approved):
Thanks for the update :)

PR #899, opened 2023-03-20
Clarify the purpose of message property if asynchronous payment cannot be initiated

saltanovas:
The change is based on the question is Stack Overflow: https://stackoverflow.com/questions/75680910/custom-error-message-if-asynchronous-payment-cant-be-initiated

Previous documentation version does not clarify how `message` is being used. Without any deeper explanation, the one may believe that it is used to show a custom error message on redirect back to the shop if asynchronous payment cannot be initiated. Even though `status` takes the effect and changes order payment status, `message` doesn't seem to make any visible effect.

Additionally, I would also like to clarify where the `message` is being logged and where those logs could be found. However, my knowledge about Shopware Payment API is very little.


mstegmeyer (approved):

guides/plugins/apps/payment.md
```diff
@@ -135,7 +135,7 @@ Keep in mind that just by providing a `message`, the payment will default to sta
...
+Shopware sends the first `pay` POST request which is supposed to start the payment with the payment provider. All necessary data is provided: the `order`, `orderTransaction` and a `returnUrl`, where the user should be redirected to once the payment process with the payment provider has been finished. If everything is correct and the payment process is ready to start, the response to this request must be a `redirectUrl`, where the user is redirected to by Shopware. In case the payment can't be started \(for example because of missing credentials for the shop\), the response can return a `fail` status and / or a `message`. However, the provided `message` will be logged only, it is not being shown on redirect back to the shop.
```
sushmangupta, line 138:
```suggestion
Shopware sends the first `pay` POST request, which is supposed to start the payment with the payment provider. All necessary data is provided: the `order`, `orderTransaction`, and a `returnUrl`, where the user should be redirected once the payment process with the payment provider has been finished. If everything is correct and the payment process is ready to start, the response to this request must be a `redirectUrl`, where the user is redirected to by Shopware. In case the payment can't be started \(for example, because of missing credentials for the shop\), the response can return a `fail` status and/or a `message`. However, the provided `message` will be logged only, not being shown on redirect back to the shop.
```

mstegmeyer: sounds good, can you apply it?

guides/plugins/apps/payment.md
```diff
@@ -135,7 +135,7 @@ Keep in mind that just by providing a `message`, the payment will default to sta
...
+Shopware sends the first `pay` POST request, which is supposed to start the payment with the payment provider. All necessary data is provided: the `order`, `orderTransaction`, and a `returnUrl`, where the user should be redirected once the payment process with the payment provider has been finished. If everything is correct and the payment process is ready to start, the response to this request must be a `redirectUrl`, where the user is redirected to by Shopware. In case the payment can't be started \(for example, because of missing credentials for the shop\), the response can return a `fail` status and/or a `message`. However, the provided `message` will be logged only, not being shown on redirect back to the shop.
```
sushmangupta, line 138: "However, the provided `message` will be logged only, not being shown on redirect back to the shop." What does this mean?
Can I convey it like this : "However, the provided `message` is a logged status and does not redirect back to the shop."

mstegmeyer: The message has nothing to do with the redirect. The message is an error message, that can be logged to the shopware logs, but automatically means, if it is transmitted, that the transaction has failed. The message is not shown to the user but only put into the logs.

Maybe something like this: If you provide a message, the payment process will fail automatically and a generic error message is shown to the user. The provided message will be shown in the log files.

sushmangupta: Yes this sounds better.

sushmangupta, line 138:
```suggestion
Shopware sends the first `pay` POST request, which is supposed to start the payment with the payment provider. All necessary data is provided: the `order`, `orderTransaction`, and a `returnUrl`, where the user should be redirected once the payment process with the payment provider has been finished. If everything is correct and the payment process is ready to start, the response to this request must be a `redirectUrl`, where the user is redirected to by Shopware. In case the payment can't be started \(for example, because of missing credentials for the shop\), the response can return a `fail` status and/or a `message`. If you provide a message, the payment process will fail automatically and a generic error message is shown to the user. The provided message will be shown in the log files.
```

sushmangupta (approved):

PR #917, opened 2023-04-06
Add path to config files were applicable

Schrank:


bojanrajh (approved):

btxtiger:
Can you please give a hint, why the path is `config/packages/prod/shopware.yaml` and not `config/packages/shopware.yaml` ?
Does it go along with APP_ENV, so possibly different configuration files for each APP_ENV can be created?

Christian-Rades:
@btxtiger Yes, the reason is to have the configs only affect the prod environment. This is default symfony behavior please see this [document](https://symfony.com/doc/current/configuration.html#configuration-environments) for details

PR #930, opened 2023-04-20
Update Admin port in devenv doc

sneakyvv:
The port for Adminer is 8010, see [the .nix file](https://github.com/shopware/recipes/blob/main/frosh/devenv-meta/0.2/root/devenv.nix#L89)


sushmangupta:
Thanks for the update! :)

PR #936, opened 2023-04-30
refactor: Remove module.hot.accept()

aragon999:
As far as I know this is not needed anymore, at least I never included it in my plugins and it is not mentioned anywhere else anymore.


tobiasberge (approved):
Hi @aragon999,

thanks for the hint.

I tested this in my env and also wasn't able to see any difference. Even when I do changes within node_modules of a plugin/app, I still have a refresh.

Therefore I would agree that we can remove this for now. Less code, less confusion.

PR #938, opened 2023-05-02
NEXT-22759 - Using custom entities in custom field selects

AydinHassan:


guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
 POST /api/search/custom-entity-blog
 ```
 
+## Custom Fields
```
sushmangupta, line 43: Shouldn't the heading be Custom Entities instead?

guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
...
+        <fields>
+            <string name="name" required="true" translatable="true" store-api-aware="true" />
+            <price name="discount" required="true" store-api-aware="true"/>
+            <many-to-many name="products" reference="product" store-api-aware="true" />
+        </fields>
+    </entity>
+</entities>
+```
+
+{% endcode %}
```
sushmangupta, line 62: The first part of this article has a code block with similar info to register the custom entities as below:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<entities xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/platform/trunk/src/Core/System/CustomEntity/Xml/entity-1.0.xsd">
    <entity name="custom_entity_bundle">
        <fields>
            <string name="name" required="true" translatable="true" store-api-aware="true" />
            <price name="discount" required="true" store-api-aware="true"/>
            <many-to-many name="products" reference="product" store-api-aware="true" />
        </fields>
    </entity>
</entities>
```
May I know if the new content is a replacement for that?

guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
...
+
+First we set `custom-fields-aware` to true. Then we specify a field to use as a label for the entity when selecting via the custom field. In this case, we use the `name` field. The field must exist in the `fields` section of the entity definition and it must be of type `string`.
```
sushmangupta, line 64:
```suggestion
First, we set `custom-fields-aware` to true. Then we specify a field to use as a label for the entity when selecting via the custom field. In this case, we use the `name` field. The field must exist in the `fields` section of the entity definition and it must be of type `string`.
```

guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
...
+
+Now you will find your entity in the "Entity Type" select when creating a custom field of type "Entity Select". Without a snippet label for the entity it will display as `custom_entity_bundle.label`. You can create a snippet to add a label like so:
```
sushmangupta, line 66:
```suggestion
Now you will find your entity in the "Entity Type" select when creating a custom field of type "Entity Select". Without a snippet label for the entity, it will display as `custom_entity_bundle.label`. You can create a snippet to add a label like so:
```

guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
 POST /api/search/custom-entity-blog
 ```
 
+## Custom Fields
+
+By default, it is not possible to create a custom field of type "Entity Select" which references a custom entity. However, you can opt in to this behavior. You will need to add the `custom-fields-aware` & `label-property` attributes to your entity definition:
```
sushmangupta, line 45:
```suggestion
By default, it is not possible to create a custom field of type "Entity Select", which references a custom entity. However, you can opt in to this behavior. You will need to add the `custom-fields-aware` & `label-property` attributes to your entity definition:
```

sushmangupta (changes requested):
@AydinHassan, please take a look at the requested changes

guides/plugins/apps/custom-data/custom-entities.md
AydinHassan: It's about using custom entities to custom fields, because you can create an custom field of "entity" type and then assign to it an entity, here we are adding support to assign custom entities as the entity type. I am open to suggestions though, maybe "Custom entities as custom fields" ?

guides/plugins/apps/custom-data/custom-entities.md
AydinHassan: Almost, we add the attributes: `custom-fields-aware="true" label-property="name"` to the `<entity` line. Is there a better way to express the changes ?

guides/plugins/apps/custom-data/custom-entities.md
sushmangupta: Okay got it 👍 
How about : "Using Custom Entities to Add Custom Fields" or "Add Custom Fields to Custom Entities"?

guides/plugins/apps/custom-data/custom-entities.md
sushmangupta: I think we can keep it as is, as shared by you because they are relevantly different topics. 👍

guides/plugins/apps/custom-data/custom-entities.md
```diff
@@ -40,6 +40,43 @@ Additionally, to the repository you can also access your custom entities via [Ad
...
+            <price name="discount" required="true" store-api-aware="true"/>
+            <many-to-many name="products" reference="product" store-api-aware="true" />
+        </fields>
+    </entity>
+</entities>
+```
+
+{% endcode %}
+
+First, we set `custom-fields-aware` to true. Then we specify a field to use as a label for the entity when selecting via the custom field. In this case, we use the `name` field. The field must exist in the `fields` section of the entity definition and it must be of type `string`.
```
sushmangupta, line 64:
```suggestion
To enable the usage of custom fields, the `custom-fields-aware` setting should be set to true. Then, it is necessary to indicate a label field for the entity that will be used when selecting via the custom field. In this example, the `name` field is selected as the `label-property` field. It is important to note that this field must be included in the `fields` section of the entity definition and be of type `string`.
```

sushmangupta (changes requested):

guides/plugins/apps/custom-data/custom-entities.md
AydinHassan: I used "Using Custom Entities with Custom Fields" I hope that is okay :)

sushmangupta (approved):

PR #940, opened 2023-05-03
There is no StoreApiClient in 6.5

KaiOnGitHub:
see https://github.com/shopware/platform/blob/trunk/UPGRADE-6.5.md#removal-of-the--_proxystore-api-api-route Would be nice to update this section with some clarity on how to work with the store api from the storefront.


JoshuaBehrens:
Is this already helpful? https://shopware.stoplight.io/docs/store-api/

KaiOnGitHub:
Actually I don't know if the store-api should be called directly from the storefront at all anymore. Maybe @shyim could confirm this. If that is the case it would be nice to mention this quickly as a reference to the old documentation and give a recommendation on how to fetch data from the storefront. 

The way I understand is that you should write your own endpoints that internally call the store-api and these you call from the storefront.

shyim (approved):
Removing that is fine. Our suggestion should be that the controllers get's duplicated for store-api and storefront. (or just annotating 2 Route attributes). See like this https://github.com/shopware/docs/pull/591

sushmangupta:
Thank you all for the contribution :)

PR #942, opened 2023-05-03
Update B2B naming conventions

sushmangupta:
This ticket is related to [DX-353](https://shopware.atlassian.net/browse/DX-353)


sushmangupta:
This is rejected as the request only applies to User guide and not dev docs.

PR #951, opened 2023-05-09
Update requirements.md

null93:
Update recommended stack to include node 18 instead of node 16. 18 is recommended above.


sushmangupta (approved):
Thanks for pointing it out! :)

PR #961, opened 2023-05-23
Don't abbreviate composer require

amenk:
... and keep it consistent with other calls in the same docs


sushmangupta (approved):

sushmangupta:
Thanks for your contribution :)

PR #964, opened 2023-05-26
Additional flag for media command

magnolia4:


magnolia4:
The folder-entity flag information should be very useful ;-)

bojanrajh (changes requested):

resources/references/core-reference/commands-reference.md
```diff
@@ -184,7 +184,7 @@ $ bin/console [command] [parameters]
...
-| `media:delete-unused`        | Deletes all media files that are never used. Use the `--dry-run` flag to see a paginated list of files that will be deleted, without actually deleting them. Use the `--grace-period-days=10` to set a grace period for unused media, meaning that only media uploaded before the current date and time minus 10 days will be considered for deletion. The default is 20 and therefore any media uploaded in the previous 20 days will not be considered for deletion even if it is unused. |
+| `media:delete-unused`        | Deletes all media files that are never used. Use the `--dry-run` flag to see a paginated list of files that will be deleted, without actually deleting them. Use the `--grace-period-days=10` to set a grace period for unused media, meaning that only media uploaded before the current date and time minus 10 days will be considered for deletion. The default is 20 and therefore any media uploaded in the previous 20 days will not be considered for deletion even if it is unused. Use the --folder-entity flag to target only a specific folder (e.g. --folder-entity=PRODUCT to purge all product images) |
```
sushmangupta, line 187:
```suggestion
| `media:delete-unused`        |  Deletes all media files that are never used. Use the `--dry-run` flag to see a paginated list of files that will be deleted, without actually deleting them. Use the `--grace-period-days=10` to set a grace period for unused media, meaning only media uploaded before the current date and time minus 10 days will be considered for deletion. The default is 20 and therefore any media uploaded in the previous 20 days will not be considered for deletion even if it is unused. Use the `--folder-entity` flag to target only a specific folder (e.g. `--folder-entity=PRODUCT` to purge all product images) |
```

PR #975, opened 2023-06-15
update link

sushmangupta:


bojanrajh (approved):
It will work, but this would be better `../../../guides/plugins/overview`.

PR #976, opened 2023-06-15
bootstap 5 update

levin192:
bootstrap 5 uses different data attributes than bs 4

### DIFF
--- a/guides/plugins/plugins/storefront/using-a-modal-window.md
+++ b/guides/plugins/plugins/storefront/using-a-modal-window.md
@@ -73,7 +73,7 @@ When setting `data-ajax-modal="true"` together with `data-url` shopware automati
 {% block base_main_inner %}
     <!-- This uses `AjaxModalPlugin` -->
     <button class="btn"
-            data-bs-toggle="modal"
+            data-ajax-modal="true"
             data-url="https://example.org/ajax-url">
 
     {{ parent() }}
--- a/guides/plugins/plugins/storefront/using-a-modal-window.md
+++ b/guides/plugins/plugins/storefront/using-a-modal-window.md
@@ -62,7 +62,7 @@ Here is a basic implementation as an example. We override the `base_main_inner` 
 
 ## Create a modal using AjaxModalPlugin
 
-When setting a `data-url` in addition to `data-bs-toggle="modal"` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -62,7 +62,7 @@ Here is a basic implementation as an example. We override the `base_main_inner`
 
 ## Create a modal using AjaxModalPlugin
 
-When setting a `data-url` in addition to `data-toggle="modal"` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:
+When setting a `data-url` in addition to `data-bs-toggle="modal"` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:
```
tobiasberge, line 65:
```suggestion
When setting `data-ajax-modal="true"` together with `data-url` shopware automatically uses the `PseudoModalUtil` and the pseudo modal template from the `base.html.twig` to render a modal:
```

We actually changed this in 6.5.0 to avoid interfering with the default Bootstrap modal opening mechanism, and also separate it more clearly as a shopware JS-plugin.

> Deprecated selector `[data-bs-toggle="modal"][data-url]` to initialize the  AjaxModal plugin. Use selector `[data-ajax-modal][data-url]` instead.

The current syntax still works until major version 6.6.0, but it produces a JS error in the console and `data-ajax-modal` is the recommended way.

Related changelog and upgrade notes: https://github.com/shopware/platform/blob/trunk/changelog/release-6-5-0-0/2023-01-04-fix-console-error-on-ajax-modal.md

guides/plugins/plugins/storefront/using-a-modal-window.md
```diff
@@ -73,7 +73,7 @@ When setting a `data-url` in addition to `data-toggle="modal"` shopware automati
 {% block base_main_inner %}
     <!-- This uses `AjaxModalPlugin` -->
     <button class="btn"
-            data-toggle="modal"
+            data-bs-toggle="modal"
```
tobiasberge, line 76:
```suggestion
            data-ajax-modal="true"
```

tobiasberge (changes requested):
Hi @levin192,

good catch and thanks for your contribution. 👍 

The first example with the regular modal is correct, I have one change request for the `AjaxModal`.

sushmangupta (approved):

tobiasberge (approved):

PR #987, opened 2023-07-04
Fix flow builder trigger version

elkmod:


sushmangupta (approved):
Thanks for updating @elkmod

PR #994, opened 2023-07-10
add .env overwrite hint

shyim:


sushmangupta:
@shyim , are there any changes to this: 
image

guides/installation/template.md
sandoba, line 1: Just a small fix: "hightly recommanded" > "highly recommended"

shyim: thanks fixed

shyim:
Yes @sushmangupta for migration part it's still necessary otherwise the default `.env` is not created

PR #995, opened 2023-07-10
NEXT-29147 - Add warning about twig use usage

ssltg:
Add a warning about the usage of the `use` tag in twig.

### DIFF
--- a/resources/references/storefront-reference/twig-function-reference.md
+++ b/resources/references/storefront-reference/twig-function-reference.md
@@ -7,7 +7,7 @@ In Shopware, we extend Twig's functionality by custom ones. See our own actions 
 
 Try to avoid importing blocks from the core templates with the \{\% use \%\} tag for horizontal reuse in twig. [Twig documentation - use tag](https://twig.symfony.com/doc/3.x/tags/use.html).
 
-The \{\% use \%\} tag does not consider the template inheritance similar to `{% sw_extends %}`.
+The \{\% use \%\} tag does not consider the template inheritance similar to \{\% sw_extends \%\}`.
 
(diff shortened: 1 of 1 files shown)


resources/references/storefront-reference/twig-function-reference.md
```diff
@@ -2,6 +2,13 @@
 
 In Shopware, we extend Twig's functionality by custom ones. See our own actions below.
 
+{% hint style="warning" %}
+**We do not support the usage of the `use` tag in twig**
+
+Be careful while using the `use` tag for horizontal reuse in twig. [Twig documentation - use tag](https://twig.symfony.com/doc/3.x/tags/use.html).
+Because it can break the inheritance we do not support the usage of the `use` tag with our `Storefront` templates. Any change in our templates can break your plugin or app.
```
tobiasberge, line 9:
```suggestion
**No official support for the twig `{% use %}` tag**

Try to avoid importing blocks from the core templates with the `{% use %}` tag for horizontal reuse in twig. [Twig documentation - use tag](https://twig.symfony.com/doc/3.x/tags/use.html).

The `{% use %}` tag does not consider the template inheritance similar to `{% sw_extends %}`.

Furthermore, templates which are imported via `{% use %}` are not allowed to have additional twig statements outside of twig blocks. Therefore, changes in core templates which are imported via `{% use %}` might break your app or plugin.
```

I tried to explain it a bit more and tried to rephrase the "using the use".

tobiasberge (changes requested):
Just added one suggestion.

tobiasberge (approved):

PR #1001, opened 2023-07-14
Clarifying sw_sanitize

quisse:
https://stackoverflow.com/questions/76676485/when-do-we-need-to-use-sw-sanitize/76676684#comment135186785_76676684


resources/references/storefront-reference/twig-function-reference.md
```diff
@@ -30,7 +30,7 @@ Furthermore, templates which are imported via \{\% use \%\} are not allowed to h
 | :--- | :--- | :--- |
 | `replace_recursive` | Enables recursive replacement in addition to twig's default `replace` filter | To see an example, see the guide on [add custom JavaScript](../../../guides/plugins/plugins/storefront/add-custom-javascript.md) |
 | `currency` | Adopts currency formatting: The currency symbol and the comma setting. | --- |
-| `sw_sanitize` | Filters tags and attributes from a given string. | --- |
+| `sw_sanitize` | Filters tags and attributes from a given string. By default twig's autoescaping is on so this filter is to explicitly allow basic html tags like <i>,<b>,... | --- |
```
sushmangupta, line 33:
```suggestion
| `sw_sanitize` | Filters tags and attributes from a given string. By default, twig's auto escaping is on, so this filter explicitly allows basic HTML tags like <i>,<b>,... | --- |
```

sushmangupta (approved):
Thanks for the updating the info :)

PR #1031, opened 2023-07-28
Update quality guidelines link

sushmangupta:
The hyperlinks are not getting redirected. Only when refreshed, it gets pointed to the right anchor.


sushmangupta:
Works fine now.

PR #1035, opened 2023-07-30
Update add-custom-rules.md

waldicom:
Typo


waldicom:
Yes, that's right. Thank you.

PR #1038, opened 2023-08-01
Update using-composer-dependencies.md

KaiOnGitHub:
Not 100% sure on this, but I thought the whole idea of the **executeComposerCommands** function was that we add the vendor folder on plugin installation and the dependencies are installed in the root vendor folder and not in the plugin's vendor folder?


KaiOnGitHub:
Okay, so the way I understand it is, that you have two options:
1. You bundle your plugin's dependencies inside the plugin's own vendor folder and add `require __DIR__ . '/../vendor/autoload.php';` in your plugin base class
2. You let Shopware handle the plugin's dependencies, remove the vendor folder and add the `executeComposerCommands ` function to your plugin base class

Option two is the recommended way, see https://www.shopware.com/en/news/the-improved-shopware-installer/
Quote: 
> You do not have to use the new method of letting Shopware take care of your dependencies. But please do so, unless you have a good reason not to. It’s very straightforward.

PR #1039, opened 2023-08-02
Update deployer guide

Isengo1989:
Updated the guide to work with the latest SW version, 6.5.3.3

image

Also 

- added some more info on the bin/build-js.sh command 
- different way how to use deployer on runner
- use env since setLabel was removed etc
- added cachetool for cleaner opcache after deployment


sushmangupta (approved):
Thanks for updating all the info @Isengo1989 :)

PR #1043, opened 2023-08-07
DX377 - Improvise custom rules examples

Isengo1989:
- Change example to a little more real e-commerce scenaria
- Change image
- Add multi-select hint


King-of-Babylon (approved):
LGTM :+1:

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -20,57 +20,76 @@ To create a custom rule, we have to implement both backend \(PHP\) code and a us
 
 ### Creating rule in PHP
 
-First of all we need a new Rule class, in this example we will name it `LunarEclipseRule`. It will be placed in the directory `<plugin root>/src/Core/Rule`. Our new class has to extend from the abstract class `Shopware\Core\Framework\Rule\Rule`. Below you can find an example implementation.
+First of all we need a new Rule class, in this example we will name it `FirstMondayOfTheMonthRule`. It will be placed in the directory `<plugin root>/src/Core/Rule`. Our new class has to extend from the abstract class `Shopware\Core\Framework\Rule\Rule`. Below you can find an example implementation.
```
sushmangupta, line 23:
```suggestion
First of all, we need a new Rule class. In this example, we name it as `FirstMondayOfTheMonthRule`. It will be placed in the directory `<plugin root>/src/Core/Rule`. Our new class has to extend from the abstract class `Shopware\Core\Framework\Rule\Rule`. Below you can find an example implementation.
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -104,26 +129,25 @@ Now we want to implement our new rule in the Administration so that we can manag
...
+As you can see, this is decorating the `RuleConditionService` by using its name `ruleConditionDataProviderService`. The decoration adds a new condition called `first_monday`. Make sure to match the name we've used in the `getName` method in PHP. Next, we define the component, in our case `swag-first-monday`, which is responsible for rendering the rule inside the Administration. We will create this component in the next step. Furthermore we defined a label, which will be displayed in the rule builder selection. The last option is the scope, which in our case is `global`, as we have not specified a specific one in our core class.
```
sushmangupta, line 150:
```suggestion
As you can see, this is decorating the `RuleConditionService` by using its name `ruleConditionDataProviderService`. The decoration adds a new condition called `first_monday`. Make sure to match the name we have used in the `getName` method in PHP. Next, we define the component, in our case, `swag-first-monday`, which is responsible for rendering the rule inside the Administration. We will create this component in the next step. Furthermore, we defined a label, which will be displayed in the rule builder selection. The last option is the scope, which in our case is `global`, as we have not specified a specific one in our core class.
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -272,7 +305,7 @@ Shopware.Application.addServiceProviderDecorator('ruleConditionDataProviderServi
 What do `notEquals` and `equalsAny` actually mean?
 With these two properties, you can define the rules you want to assign to a specific relation, i.e., `productPrices` need to have at least one condition inside `equalsAny` or should not have any condition inside of `notEquals`.
 
-Finally, you just need a new snippet. With that said, you successfully defined restrictions for your custom condition. If you now try to assign a rule with your condition to advanced prices, you should see that it is not possible, and the rule is disabled.
+Finally, you just need a snippet, you can choose an existing one or create one yourself. With that said, you successfully defined restrictions for your custom condition. If you now try to assign a rule with your condition to advanced prices, you should see that it is not possible, and the rule is disabled.
```
sushmangupta, line 308:
```suggestion
Finally, you just need a snippet, and you can choose an existing one or create one yourself. With that said, you successfully defined restrictions for your custom condition. If you now try to assign a rule with your condition to advanced prices, you should see that it is not possible, and the rule is disabled.
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -214,7 +238,7 @@ The last step is, creating a template for our condition. We will create a new fi
 {% endraw %}
 {% endcode %}
 
-As you can see, our `sw-single-select` uses the previously created computed property `selectValues` as the `options` prop, and the value is saved into the variable `isLunarEclipse`. That's it, your rule is now fully integrated.
+As you can see, our `sw-single-select` uses the previously created computed property `selectValues` as the `options` prop, and the value is saved into the variable `isFirstMondayOfTheMonth`. That's it, your rule is now fully integrated.
```
sushmangupta, line 241:
```suggestion
As you can see, our `sw-single-select` uses the previously created computed property `selectValues` as the `options` prop, and the value is saved into the variable `isFirstMondayOfTheMonth`. That's it; your rule is now fully integrated.
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -188,24 +212,24 @@ Shopware.Component.extend('swag-lunar-eclipse', 'sw-condition-base', {
...
+As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call this.$tc\('global.sw-condition.condition.yes'\), it's just loading a translation by its name, in this case "Yes" and "No". Note: When dealing with boolean values, make sure to always return strings here!
```
sushmangupta, line 215:
```suggestion
As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call this.$tc\('global.sw-condition.condition.yes'\); it's just loading a translation by its name, in this case, "Yes" and "No". 
{%hint style = "info"%} When dealing with boolean values, make sure to always return strings.{%endhint%}
```

sushmangupta (changes requested):

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -188,24 +212,25 @@ Shopware.Component.extend('swag-lunar-eclipse', 'sw-condition-base', {
...
+As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call this.$tc\('global.sw-condition.condition.yes'\); it's just loading a translation by its name, in this case, "Yes" and "No".
```
sushmangupta, line 215:
```suggestion
As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call `this.$tc\('global.sw-condition.condition.yes'\)`; it's just loading a translation by its name, in this case, "Yes" and "No". 
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -188,24 +212,25 @@ Shopware.Component.extend('swag-lunar-eclipse', 'sw-condition-base', {
...
+As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call `this.$tc\('global.sw-condition.condition.yes'\)`; it's just loading a translation by its name, in this case, "Yes" and "No".
```
sushmangupta, line 215:
```suggestion
As you can see, our `swag-first-monday` has to extend from the `sw-condition-base` component and has to bring a custom template, which will be explained in the next step. Let's have a look at each property and method. The first computed property is `selectValues`, which returns an array containing the values "true" and "false". Those will be used in the template later on, as they will be the selectable options for the shop administrator. Do not get confused by the call `this.$tc\('global.sw-condition.condition.yes'\)`; it's just loading a translation by its name, in this case, "Yes" and "No".
```

sushmangupta (approved):

sushmangupta:
@Isengo1989 , thank you for updating the info : )

PR #1065, opened 2023-08-11
Fix order in deployer docs

Isengo1989:


sushmangupta (approved):
Thank you for updating @Isengo1989

PR #1067, opened 2023-08-11
Add plugin list with available infos

Isengo1989:
Adding all plugins that extend from `plugin` to get an overview for developers.

A lot of info is missing, the question would be are we adding this or do we plan to replace most of them anyways mid-longterm?


sushmangupta:
@King-of-Babylon , Can you please take a look at this PR ?

King-of-Babylon:
> @King-of-Babylon , Can you please take a look at this PR ?

Hi @sushmangupta,
maybe @tobiasberge can help you with this :blush:

PR #1081, opened 2023-08-25
Extending services Code Example Update

GerDner:
Extends Code example to not break admin ui.

"name" param should always be defined? Example will break the admin ui in certain modules because data is missing. In our case, permission view was empty because all permission the resettet.


jleifeld (approved):

sushmangupta:
@GerDner , Thanks for updating the docs. :)

PR #1085, opened 2023-08-28
Update PHPunit docs

Isengo1989:
Update docs to describe:

- new [plugin command file generation](https://github.com/shopware/platform/commit/266102c72cc149c276a166d8ea8e46c0f4370e66)
- flex template info
- .dist removal

**Dependency**: https://github.com/shopware/dev-tools/pulls

### DIFF
--- a/guides/plugins/plugins/testing/php-unit.md
+++ b/guides/plugins/plugins/testing/php-unit.md
@@ -14,7 +14,7 @@ Furthermore, you should have a look at our [Execute database queries/migrations]
 
 ## PHPUnit configuration
 
-First of all we have to configure PHPUnit a bit. Therefore we have to create a file called `phpunit.xml` in the root directory of our plugin. To get more familiar with the configurable options, you can refer to the [PHPUnit documentation](https://phpunit.readthedocs.io/en/8.5/configuration.html). In this example we configure PHPUnit to search in the directories `<plugin root>/src/Test` and `<plugin root>/src/Migration/Test` for our tests.
+First, to configure PHPUnit, create a file called `phpunit.xml` in the root directory of your plugin. To get more familiar with the configurable options, refer to the [PHPUnit documentation](https://phpunit.readthedocs.io/en/8.5/configuration.html). This example explains configuring PHPUnit to search in the directories `<plugin root>/src/Test` and `<plugin root>/src/Migration/Test` for your tests.
 
 The file below is automatically generated when using the `bin/console plugin:create` command
 
--- a/guides/plugins/plugins/testing/php-unit.md
+++ b/guides/plugins/plugins/testing/php-unit.md
@@ -6,10 +6,6 @@ In this guide you'll learn how to create PHPUnit tests in Shopware 6. You can re
 <!-- markdown-link-check-disable-next-line -->
 {% embed url="https://phpunit.de/documentation.html" caption="" %}
 
-{% hint style="info" %}
-Throughout this guide, you will find the `$` symbol representing your command line.
-{% endhint %}
-
 ## Prerequisites
 
 In order to create tests for your plugin, you first need a plugin as base. Therefore, you can refer to the [Plugin Base Guide](../plugin-base-guide.md).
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/testing/php-unit.md
```diff
@@ -196,72 +187,33 @@ cd ./../../../
 Now we can execute the `phpunit.sh` to run all tests of our plugin.
 
 ```text
-$ ./bin/phpunit.sh
+./bin/phpunit.sh
```
shyim, line 190: do we really need this wrapper script? I would rather show the real phpunit command with the options. Feels weird to wrap it especially in the docs.

Isengo1989: I guess it was added to remove one point of error since passing the wrong path will break the execution. But to understand the command better, adding the full command makes sense :+1:

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -196,72 +187,33 @@ cd ./../../../
...
 ```
 
 ### Executing a single test
 
 If we want to execute a specific test only, we have to pass the path to the test as argument.
 
 ```text
```
JoshuaBehrens, line 197: If you remove the leading `$` as indicator for a user/non-root command, wouldn't it be a good idea to mark these sections as shell scripts so highlighting suggest, what it could be instead?

Isengo1989: @JoshuaBehrens yes, they will be marked as ```shell in the future.

We discussed this in the team and the benefit of easy copying overweighed the indicator benefit. In 99% of the cases the usage is clear (bin/console etc.)

JoshuaBehrens: I am with you. It is really annoying to remove these leading indicators :) just wanted to know, whether you will update the codefence hint as well

JoshuaBehrens (approved):

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -18,46 +14,41 @@ Furthermore, you should have a look at our [Execute database queries/migrations]
 
 ## PHPUnit configuration
 
-First of all we have to configure PHPUnit a bit. Therefore we have to create a file called `phpunit.xml.dist` in the root directory of our plugin. To get more familiar with the configurable options, you can refer to the [PHPUnit documentation](https://phpunit.readthedocs.io/en/8.5/configuration.html). In this example we configure PHPUnit to search in the directories `<plugin root>/src/Test` and `<plugin root>/src/Migration/Test` for our tests.
+First of all we have to configure PHPUnit a bit. Therefore we have to create a file called `phpunit.xml` in the root directory of our plugin. To get more familiar with the configurable options, you can refer to the [PHPUnit documentation](https://phpunit.readthedocs.io/en/8.5/configuration.html). In this example we configure PHPUnit to search in the directories `<plugin root>/src/Test` and `<plugin root>/src/Migration/Test` for our tests.
```
sushmangupta, line 17:
```suggestion
First, to configure PHPUnit, create a file called `phpunit.xml` in the root directory of your plugin. To get more familiar with the configurable options, refer to the [PHPUnit documentation](https://phpunit.readthedocs.io/en/8.5/configuration.html). This example explains configuring PHPUnit to search in the directories `<plugin root>/src/Test` and `<plugin root>/src/Migration/Test` for your tests.
```

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -18,46 +14,41 @@ Furthermore, you should have a look at our [Execute database queries/migrations]
...
 
-Here's an example configuration for the development template:
+The file below is automatically generated when using the `bin/console plugin:create` command
```
sushmangupta, line 19:
```suggestion
The file below is automatically generated when using the `bin/console plugin:create` command.
```

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -68,7 +59,7 @@ Here's an example configuration for the development template:
 
 After we've configured PHPUnit, we can start writing our first test. In this example we have a test which simply tries to instantiate every `.php` class, to see if any used core classes went missing. In our test we use the `IntegrationTestBehaviour` trait which comes in with some handy features, such as automatically setting up a database transaction or clearing the cache before starting your tests.
```
sushmangupta, line 60:
```suggestion
After having configured PHPUnit, you can start writing your first test. In this example, you have a test that simply tries to instantiate every `.php` class to see if any used core classes are missing. In your test, you use the `IntegrationTestBehaviour` trait, which comes with some handy features, such as automatically setting up a database transaction or clearing the cache before starting your tests.
```

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -115,9 +106,9 @@ class UsedClassesAvailableTest extends TestCase
 
 ### Migration test
 
-In order to test our example migration `Migration1611740369ExampleDescription`, we create a new test called `Migration1611740369ExampleDescriptionTest` which extends from the PHPUnit `TestCase`. Furthermore we use the `KernelTestBehaviour` trait since we need our database connection from the container.
+In order to test our example migration `Migration1611740369ExampleDescription`, we create a new test called `Migration1611740369ExampleDescriptionTest` which extends from the PHPUnit `TestCase`. We use the `KernelTestBehaviour` trait since we need our database connection from the container.
```
sushmangupta, line 109:
```suggestion
In order to test the example migration `Migration1611740369ExampleDescription`, create a new test called `Migration1611740369ExampleDescriptionTest`, which extends from the PHPUnit `TestCase`. Use the `KernelTestBehaviour` trait since you need the database connection from the container.
```

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -175,93 +166,41 @@ In some cases you want a service to behave differently in the test run. Such a c
...
-Now we can execute the `phpunit.sh` to run all tests of our plugin.
-
-```text
-$ ./bin/phpunit.sh
-```
-
-### Executing a single test
-
-If we want to execute a specific test only, we have to pass the path to the test as argument.
+If no testsuite is passed it will execute all testsuites.
```
sushmangupta, line 181:
```suggestion
If no testsuite is passed, it will execute all testsuites.
```

sushmangupta (changes requested):

guides/plugins/plugins/testing/php-unit.md
```diff
@@ -2,62 +2,53 @@
...
 {% embed url="https://phpunit.de/documentation.html" caption="" %}
 
-{% hint style="info" %}
-Throughout this guide, you will find the `$` symbol representing your command line.
-{% endhint %}
-
 ## Prerequisites
 
-In order to create tests for your plugin, you first need a plugin as base. Therefore, you can refer to the [Plugin Base Guide](../plugin-base-guide.md).
+In order to create tests for a plugin, you need a plugin as a base. Please refer to the [Plugin Base Guide](../plugin-base-guide.md).
```
sushmangupta, line 11:
```suggestion
In order to create tests for a plugin, you need a plugin as a base. Refer to the [Plugin Base Guide](../plugin-base-guide.md) for more information.
```

sushmangupta (changes requested):

sushmangupta (approved):
@Isengo1989 , thanks for updating the info 👍

PR #1086, opened 2023-08-28
feat: add scheduled task page

shyim:


resources/references/core-reference/commands-reference.md
```diff
@@ -251,10 +251,12 @@ $ bin/console [command] [parameters]
 
 ### Scheduled-task
 
-| Command | Description |
+| Command | Description | Version
```
Isengo1989, line 254: @sushmangupta @shyim what do you think about adding a version to the command references in the future?

Isengo1989 (approved):

resources/references/core-reference/commands-reference.md
shyim: sounds good

PR #1121, opened 2023-09-20
Add mermaid flowcharts

Isengo1989:
image

@bojanrajh is there a way how to zoom into mermaid or add a modal to show the full width?


bojanrajh:
> @bojanrajh is there a way how to zoom into mermaid or add a modal to show the full width?

Not at the moment. I'll create a ticket for adding the functionality, maybe something like [this](https://github.com/vuejs/vitepress/issues/854) will work.

Isengo1989:
> > @bojanrajh is there a way how to zoom into mermaid or add a modal to show the full width?
> 
> Not at the moment. I'll create a ticket for adding the functionality, maybe something like [this](https://github.com/vuejs/vitepress/issues/854) will work.

Ok, that would work :+1: 

So I guess the function did not make it yet into the core of vitepress but we can use a component like it is described in the  PR, right?

Isengo1989:
**Update**: This PR is currently on hold till we get a solution for SVG zoom.

PR #1124, opened 2023-09-26
Add media

Isengo1989:


Isengo1989:
@bojanrajh did you check about the twig syntax? I found a few that already are included like that, so I guess it is not a problem

image

bojanrajh:
@Isengo1989 Seems like it was fixed, so we can safely use twig syntax inside the twig codeblock. For reference, I'll also revert [this](https://github.com/shopware/docs/commit/d12f07f9ea94989f45c38cb30d67794dfd354cbb#diff-0b8b029edaecdcc11bf6175daa82e7ba9491a4b4bc466418242b26748fb9f2fdL85) change.

bojanrajh (approved):

PR #1131, opened 2023-10-04
Fix json ending and add override info

Isengo1989:
- Add info so it is clear how to override default sw/bootstrap values / colors.
- Fix json ending

### DIFF
--- a/guides/plugins/plugins/storefront/add-custom-styling.md
+++ b/guides/plugins/plugins/storefront/add-custom-styling.md
@@ -58,7 +58,7 @@ body {
 This comes with the advantage that when you want to change this color for all occurrences, you only have to change this variable once and the hard coded values are not cluttered all over the codebase.
 
 ::: info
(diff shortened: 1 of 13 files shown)


guides/plugins/plugins/storefront/add-custom-styling.md
```diff
@@ -57,6 +57,10 @@ body {
 
 This comes with the advantage that when you want to change this color for all occurrences, you only have to change this variable once and the hard coded values are not cluttered all over the codebase.
 
+::: info
+Refer to the theme guide **[Override Bootstrap Variables in a Theme](../../themes/override-bootstrap-variables-in-a-theme.md)** if you want to override some of the default Shopware variables.
```
bojanrajh, line 61: Please, replace `.md` with `.html`.

bojanrajh (changes requested):
Requested one change, then we can merge it.

guides/plugins/plugins/storefront/add-custom-styling.md
Isengo1989: @bojanrajh done

PR #1142, opened 2023-10-16
Fix typo in https://developer.shopware.com/docs/guides/plugins/plugins/storefront/customize-templates.html

Kudze:
You have a typo in this page https://developer.shopware.com/docs/guides/plugins/plugins/storefront/customize-templates.html:
image

Just replaced with `{{ dump() }}`.


Isengo1989 (approved):
Thanks @Kudze for your PR :tada:

PR #1144, opened 2023-10-17
NEXT-30950 - Add a guide about how to log with monolog

AydinHassan:


guides/plugins/plugins/plugin-fundamentals/logging.md
```diff
@@ -0,0 +1,92 @@
...
+
+As a plugin developer, you may want to log certain actions or errors to a log file to aid in debugging or to simply keep a record of performed actions.
+
+## Prerequisites
+
+This guide is built upon our [plugin base guide](../plugin-base-guide.md), which explains the basics of a plugin as a whole. Make sure to have a look at it to get started on building your first plugin.
+
+## Configuring Monolog
+
+First, you must make sure that your plugin loads package configuration from the `/Resources/config/packages` folder:
```
keulinho, line 13: Maybe add a link to the symfony docs, to also make clear that this boilerplate is not expected by us, but is plain symfony stuff?

AydinHassan: yep good idea

guides/plugins/plugins/plugin-fundamentals/logging.md
```diff
@@ -0,0 +1,92 @@
+# Logging
+
+## Overview
+
+As a plugin developer, you may want to log certain actions or errors to a log file to aid in debugging or to simply keep a record of performed actions.
+
+## Prerequisites
+
+This guide is built upon our [plugin base guide](../plugin-base-guide.md), which explains the basics of a plugin as a whole. Make sure to have a look at it to get started on building your first plugin.
```
bojanrajh, line 9: Please update the link according to [vitepress](https://vitepress.dev/guide/routing#linking-between-pages).

guides/plugins/plugins/plugin-fundamentals/logging.md
```diff
@@ -0,0 +1,92 @@
...
+
+## Configuring Monolog
+
+First, you must make sure that your plugin loads package configuration from the `/Resources/config/packages` folder:
+
+{% code title="<plugin root>/src/SwagBasicExample.php" %}
```
bojanrajh, line 15: Please use [code groups](https://vitepress.dev/guide/markdown#code-groups) - old Gitbook syntax is not supported anymore.

bojanrajh (changes requested):

guides/plugins/plugins/plugin-fundamentals/logging.md
AydinHassan: nice thanks, will update

JoshuaBehrens:
Will this abandon the factory class? Although I like the idea to not use my factory class, that I introduced like 3 or 4 years ago ^^' , I am not sure whether I like to do all of this. Could we make this a compiler pass or an extension, that one load and configure in like two calls?

guides/plugins/plugins/plugin-fundamentals/logging.md
AydinHassan: Done

guides/plugins/plugins/plugin-fundamentals/logging.md
AydinHassan: Also done!

AydinHassan:
@JoshuaBehrens yes we will deprecate the Logger Factory. This is default Symfony behaviour, I'm not really sure we need to simplify it, but of course we can revisit if it becomes cumbersome.

JoshuaBehrens:
@AydinHassan I like the approach of ditching that. My assumption is, that everyone will just use the logger service as it will work and is much simpler and we all end up with one big log file/stream

AydinHassan:
@JoshuaBehrens as long as people create a dedicated channel, which you can see is like 2 loc, then the project owner can easily redirect different channels to different handlers or just discard altogether, it's pretty robust.

JoshuaBehrens:
I am not sure. I see so much code, that is more than 2 lines. Do you expect others to always add the loading of the Resources/config folder?

AydinHassan:
@JoshuaBehrens yes but that part is not specific to logging, that's just how you load any config in a Symfony Bundle, you can make it less code by just using the yaml loader and only loading directly the `monolog.yaml` file - this is just a suggestion to allow your bundle to provide any package config.

JoshuaBehrens:
All the boilerplate code :/

bojanrajh (approved):

PR #1149, opened 2023-10-24
Add docs how to add script tags to Storefront HTML

tobiasberge:
Currently, we are only documenting how to add JS to our webpack build (main.js entry).

However, there might be situations where you want to add an extra `<script>` tag to the template.

There is no explicit guide for this, it is assumed you can anticipate when knowing template extensions.


Isengo1989 (approved):
Thanks for you PR @tobiasberge

PR #1168, opened 2023-11-06
Update add-custom-controller.md

daslicht:
added missing ```method="setTwig"``` call

### DIFF
--- a/guides/plugins/plugins/storefront/add-custom-controller.md
+++ b/guides/plugins/plugins/storefront/add-custom-controller.md
@@ -130,6 +130,9 @@ Next, we need to register our controller in the DI-container and make it public.
             <call method="setContainer">
                 <argument type="service" id="service_container"/>
             </call>
+            <call method="setTwig">
+                <argument type="service" id="twig"/>
+            </call>
         </service>
     </services>
 </container>


Isengo1989:
For future reference

## Storefront Controller need to have Twig injected in future versions

The `twig` service will be private with upcoming Symfony 6.0. To resolve this deprecation, a new method `setTwig` was added to the `StorefrontController`.
All controllers which extends from `StorefrontController` need to call this method in the dependency injection definition file (services.xml) to set the Twig instance.
The controllers will work like before until the Symfony 6.0 update will be done, but they will create a deprecation message on each usage.
Below is an example how to add a method call for the service using the XML definition.

### Before

```xml
<service id="Shopware\Storefront\Controller\AccountPaymentController">
    <call method="setContainer">
        <argument type="service" id="service_container"/>
    </call>
</service>
```

### After

```xml
<service id="Shopware\Storefront\Controller\AccountPaymentController">
    <call method="setContainer">
        <argument type="service" id="service_container"/>
    </call>
    <call method="setTwig">
        <argument type="service" id="twig"/>
    </call>
</service>
```

Isengo1989:
@daslicht could you add it in [# Add Dynamic Content via AJAX Calls](https://developer.shopware.com/docs/guides/plugins/plugins/storefront/add-dynamic-content-via-ajax-calls.html#add-dynamic-content-via-ajax-calls) as well, so we got it covered for the future :slightly_smiling_face:

Isengo1989 (changes requested):
Please add info to other [example](https://developer.shopware.com/docs/guides/plugins/plugins/storefront/add-dynamic-content-via-ajax-calls.html#add-dynamic-content-via-ajax-calls)

daslicht:
> Please add info to other [example](https://developer.shopware.com/docs/guides/plugins/plugins/storefront/add-dynamic-content-via-ajax-calls.html#add-dynamic-content-via-ajax-calls)

created pull request here https://github.com/shopware/docs/pull/1178

Isengo1989:
@daslicht ok, that work's as well :slightly_smiling_face: 

Thanks for the PR :tada:

PR #1173, opened 2023-11-07
DX-410 - Addition of custom templates via app

sushmangupta:


guides/plugins/apps/customize-templates.md
```diff
@@ -0,0 +1,103 @@
...
+./bin/console cache:clear
+```
+
+::: info
+Also remember to not only activate your plugin but also to assign your theme to the correct sales channel by clicking on it in the sidebar, going to the tab Theme and selecting your theme.
+:::
+
+### Finding variables
+
+Of course this example is very simplified and does not use any variables, even though you most likely want to do that. Using variables is exactly the same like in [Twig](https://twig.symfony.com/doc/3.x/templates.html#variables) in general, so this won't be explained here in detail. Still, this is how you use a variable: `{{ variableName }}`
```
bojanrajh, line 82: The `{{ variableName }}` might require wrapping it with `<code v-pre`.

sushmangupta:
@Isengo1989 , FYI - 'FroshDevelopmentHelper' also work for apps

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+  title: Customize templates
+  position: 10
+
+---
+
+# Customize Templates
+
+## Overview
+
+This guide will cover customizing Storefront templates using app.
```
Isengo1989, line 12:
```suggestion
This guide will cover customizing Storefront templates using an app.
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
+---
+nav:
+  title: Customize templates
+  position: 10
+
+---
+
+# Customize Templates
```
Isengo1989, line 8: Is the uppcase in h1 heading wanted? I saw it is mixed in our docs

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+Before you begin, make sure you have:
+
+* A basic understanding of [Shopware app development](../app-base-guide).
+* Familiarity with the [Twig template](https://twig.symfony.com/) is beneficial.
+
+## Getting started
+
+This guide assumes you have already set up your Shopware app. If not, refer to the [app base guide](../app-base-guide) for the initial setup.
+
+The following sections gives you a very short example on how you can extend a storefront block. For simplicity sake, only the page logo is replaced with a 'Hello world!' text.
```
Isengo1989, line 25:
```suggestion
The following sections give you a very short example of how you can extend a storefront block. For simplicity's sake, only the page logo is replaced with a 'Hello world!' text.
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+
+### Setting up app's view directory
+
+First of all, in your app's root, register your app's own view path, which basically represents a path in which Shopware 6 is looking for template-files. By default, Shopware 6 is looking for a directory called `views` in your app's `Resources` directory, so the path could look like this: `<app root>/src/Resources/views`
+
+### Locating the template
+
+As mentioned earlier, this guide is only trying to replace the 'demo' logo with a 'Hello world!' text. In order to find the proper template, you can simply search for the term 'logo' inside of the `<shopware root>/src/Storefront` directory. This will eventually lead you to [this file](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Storefront/Resources/views/storefront/layout/header/logo.html.twig).
```
Isengo1989, line 33:
```suggestion
As mentioned earlier, this guide is only trying to replace the 'demo' logo with a 'Hello world!' text. In order to find the proper template, you can simply search for the term 'logo' inside the `<shopware root>/src/Storefront` directory. This will eventually lead you to [this file](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Storefront/Resources/views/storefront/layout/header/logo.html.twig).
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+
+::: info
+There's a plugin out there called [FroshDevelopmentHelper](https://github.com/FriendsOfShopware/FroshDevelopmentHelper), that adds hints about template blocks and includes into the rendered HTML. This way it's easier to actually find the proper template.
```
Isengo1989, line 36:
```suggestion
There's a plugin out there called [FroshDevelopmentHelper](https://github.com/FriendsOfShopware/FroshDevelopmentHelper), that adds hints about template blocks and includes into the rendered HTML. This way, it's easier to actually find the proper template.
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+
+Put this line at the very beginning of your file:
+
+```twig
+{% sw_extends '@Storefront/storefront/layout/header/logo.html.twig' %}
+```
+
+This is simply extending the `logo.html.twig` file from the Storefront bundle. If you would leave the file like that, it wouldn't change anything, as you are currently just extending from the original file with no overrides.
+
+To replace the logo with some custom text, take a look at the block called `layout_header_logo_link` in the original file. Its contents creates an anchor tag, which is not necessary for our case anymore, so this seems to be a great block to override.
```
Isengo1989, line 61:
```suggestion
To replace the logo with some custom text, take a look at the block called `layout_header_logo_link` in the original file. Its contents create an anchor tag, which is not necessary for our case anymore, so this seems to be a great block to override.
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+./bin/console cache:clear
+```
+
+::: info
+Also remember to not only activate your app but also to assign your theme to the correct sales channel by clicking on it in the sidebar, going to the tab Theme and selecting your theme.
+:::
+
+### Finding variables
+
+Of course this example is very simplified and does not use any variables, even though you most likely want to do that. Using variables is exactly the same like in [Twig](https://twig.symfony.com/doc/3.x/templates.html#variables) in general, so this won't be explained here in detail. However, this is how you use a variable: <code v-pre>{{ variableName }}</code>
```
Isengo1989, line 87:
```suggestion
Of course, this example is very simplified and does not use any variables, even though you most likely want to do that. Using variables is exactly the same as in [Twig](https://twig.symfony.com/doc/3.x/templates.html#variables) in general, so this won't be explained here in detail. However, this is how you use a variable: <code v-pre>{{ variableName }}</code>
```

guides/plugins/apps/storefront/customize-templates.md
```diff
@@ -0,0 +1,99 @@
...
+
+But, how do you know which variables are available to use? For this, you can just dump all available variables:
```
Isengo1989, line 89:
```suggestion
But how do you know which variables are available to use? For this, you can just dump all available variables:
```

Isengo1989 (changes requested):
@sushmangupta thx for the PR. I added some grammar fixes that we might also want to apply on the plugins docs.

guides/plugins/apps/storefront/customize-templates.md
sushmangupta: H1 is always in Camel case and rest other heading followed by that is sentence case.

Isengo1989 (approved):
:tada:

PR #1176, opened 2023-11-08
fix bash syntax hint duplication typo

nmokkenstorm:
I noticed an error on the [live version](https://developer.shopware.com/docs/guides/hosting/infrastructure/message-queue.html#execution-methods) of the queue worker documentation and figured I'd fix it


Isengo1989 (approved):
@nmokkenstorm thanks for the fix :tada:

PR #1178, opened 2023-11-09
Update add-dynamic-content-via-ajax-calls.md

daslicht:
The twig service will be private with upcoming Symfony 6.0. To resolve this deprecation, a new method setTwig was added to the StorefrontController. All controllers which extends from StorefrontController need to call this method in the dependency injection definition file (services.xml) to set the Twig instance. The controllers will work like before until the Symfony 6.0 update will be done, but they will create a deprecation message on each usage. Below is an example how to add a method call for the service using the XML definition.


Isengo1989 (approved):
@daslicht thx for your PR :tada:

daslicht:
> @daslicht thx for your PR 🎉

my pleasure, you are welcome <3

PR #1182, opened 2023-11-20
Make sure people understand this is NOT for admin

Schrank:
I came to this page via search and it took a moment to realise it is not for admin. Trying to make it harder to overview 😅


sushmangupta (approved):
Okay thank you!

PR #1205, opened 2023-11-28
Add demo file and checks

Isengo1989:
**Please see**

File ending fail:
https://github.com/shopware/docs/actions/runs/7021816331/job/19104759756

File name fail:
https://github.com/shopware/docs/actions/runs/7021842518/job/19104843087?pr=1205

As for the alternative text rule, this was always applicable, the rule was set to true with no failure.


.github/workflows/file-format-check.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  Filenamecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - shell: bash
+        run: |
+          find assets | grep "\.[A-Z]" && echo "Upper case characters found in file endings" && exit 1
+          find assets -type f -mtime -14 -printf '%f\n' | grep "[_+ ]" && echo "Underscore, plus or space found in file names" && exit 1
```
sushmangupta:
```suggestion
          find assets -type f -printf '%f\n' | grep "[^a-zA-Z0-9-]" && echo "No special characters, except hyphen, are allowed in file names" && exit 1
```

.github/workflows/file-format-check.yml
```diff
@@ -0,0 +1,13 @@
...
+on: pull_request
+
+jobs:
+  Filenamecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - shell: bash
+        run: |
+          find assets | grep "\.[A-Z]" && echo "Upper case characters found in file endings" && exit 1
```
sushmangupta, line 11:
```suggestion
          find assets -type f -printf '%f\n' | grep "\.[A-Z]" && echo "Ensure file extensions are in lowercase." && exit 1
```

.github/workflows/file-format-check.yml
Isengo1989: @sushmangupta this regex won't work since it also triggers the . in the file name. Feel free to add more special chars to the existing regex.

image

.github/workflows/file-format-check.yml
Isengo1989: -type f -printf '%f\n' is not needed here since it only checks everything after the dot

.github/workflows/file-format-check.yml
sushmangupta: how about including a dot here '[^a-zA-Z0-9-.]' ?

.github/workflows/file-format-check.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  Filenamecheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - shell: bash
+        run: |
+          find assets | grep "\.[A-Z]" && echo "ERROR: Ensure file extensions are in lowercase" && exit 1
+          find assets -type f -printf '%f\n' | grep "[_+ ]" && echo "ERROR: No special characters, except hyphen, are allowed in file names" && exit 1
```
sushmangupta, line 12:
```suggestion
          find assets -type f -printf '%f\n' | grep "[_+&/ ]" && echo "ERROR: No special characters, except hyphen, are allowed in file names" && exit 1
```

sushmangupta (approved):

PR #1229, opened 2023-12-08
NEXT-30261 - Document new reverse proxy config options

keulinho:
The location of the reverse proxy changed

### DIFF
--- a/guides/hosting/infrastructure/reverse-http-cache.md
+++ b/guides/hosting/infrastructure/reverse-http-cache.md
@@ -31,19 +31,26 @@ This setup is compatible from Shopware version 6.4.
 First, we need to activate the reverse proxy support in Shopware. To enable it, we need to create a new file in `config/packages/storefront.yaml`:
 
 ```yaml
-storefront:
-    reverse_proxy:
-        enabled: true
-        ban_method: "BAN"
-        # This needs to point to your varnish hosts
-        hosts: [ "http://varnish" ]
-        # Max parallel invalidations at same time for a single worker
-        max_parallel_invalidations: 3
-        # Redis Storage for the HTTP cache tags
-        redis_url: "redis://redis"
+shopware:
+    http_cache:
+        reverse_proxy:
+            enabled: true
+            ban_method: "BAN"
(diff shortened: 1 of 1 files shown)


guides/hosting/infrastructure/reverse-http-cache.md
```diff
@@ -31,20 +31,27 @@ This setup is compatible from Shopware version 6.4.
...
+            # Max parallel invalidations at same time for a single worker
+            max_parallel_invalidations: 3
+            # Redis Storage for the HTTP cache tags
+            redis_url: "redis://redis"
 ```
 
 Also set `SHOPWARE_HTTP_CACHE_ENABLED=1` in your `.env` file.
 
+::: info
+The configuration key changed from `storefront.reverse_proxy` up to Shopware 6.5.x to `shopware.http_cache.reverse_proxy` starting with Shopware 6.6.0.0.
```
keulinho, line 50: this might be misleading as the config looks only slightly different, therefore i added the info here

What do you think, is it needed / helpful?

guides/hosting/infrastructure/reverse-http-cache.md
```diff
@@ -31,20 +31,27 @@ This setup is compatible from Shopware version 6.4.
...
+So you will need to adjust your config while upgrading.
+If you look for the old documentation and examples, you can find it [here](https://developer.shopware.com/docs/v6.5/guides/hosting/infrastructure/reverse-http-cache.html)
```
keulinho, line 52: right now the link is broken, but it should be the correct one when this branch is the dedault

guides/hosting/infrastructure/reverse-http-cache.md
Isengo1989: I think it is a good idea, often those small changes are overlooked

Isengo1989 (approved):
Thanks for the PR :tada: 

Commit reference: https://github.com/shopware/shopware/commit/9fefbca168d14b056375215ca586b69e21be377b

shyim (approved):

PR #1232, opened 2023-12-14
Fix missing quotation in path

72haroslaw:


sushmangupta (approved):
Thanks for updating the info! :)

PR #1234, opened 2023-12-15
Fix wrong composer require command for fastly configuration in PaaS

flohessling:
The PaaS docs show the wrong `composer require` command to add the fastly flex recipe.


sushmangupta:
@flohessling , thanks for updating the info! :)

PR #1239, opened 2023-12-19
Add info about logging

Isengo1989:


JoshuaBehrens:
Can we have a note on how to redirect loggers from plugins to other logging targets? AFAIK is this the sole reason, why we shall not use the logger factory anymore.

Isengo1989:
> Can we have a note on how to redirect loggers from plugins to other logging targets? AFAIK is this the sole reason, why we shall not use the logger factory anymore. Looking at it from the perspective of a composer project owner, that needs to change things according to the hosting

I bet we can, but probably not in the scope of this PR. Can you link some info about it (I remember you had a discussion somewhere about the removal of the LoggerFactory)?

In the meantime I will create a internal ticket on the matter.

shyim (approved):

JoshuaBehrens:
Yes, the feedback I got from reviving a SW5 service xD https://github.com/shopware/shopware/pull/2766 was in this pull request. And eventually your response was in NEXT-30950 to deprecate the logger factory but how do you now make use of this as a project code owner?

PR #1243, opened 2023-12-22
Add info for es:reset

Isengo1989:
https://shopware.atlassian.net/browse/DX-287


sushmangupta (approved):
Thanks for the update! :)

PR #1248, opened 2024-01-08
add partial data loading

shyim:
https://shopware.atlassian.net/browse/DX-493


guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+## Default fields
+
+Some fields are always loaded like the `id` or join relevant fields like foreign keys, these are necessary for the API to work correctly and can't be removed.
+
+## Runtime fields
+
+Some fields in the API are generated at runtime like `isSystemDefault` of the currency. These fields are loaded by default when the depending on data is available, otherwise they can be referenced in the `fields` parameter to force the API to load them.
```
sushmangupta, line 107: what is meant by "when the depending on data" here ?

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+
+For custom entity definitions with Runtime flag, the depending on fields needs to be specified inside the constructor. See an example from the core:
```
sushmangupta, line 109:
```suggestion
For custom entity definitions with runtime flag, the depending on fields needs to be specified inside the constructor. See an example from the core:
```

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+
+        // When this field is requested, we need the data of path field to generate the url
+        (new StringField('url', 'url'))->addFlags(new ApiAware(), new Runtime(['path'])),
+    ]);
+}
+```
+
+## Limitations
+
+The current limitation of the Partial Data loading is that it only works on the Entity level. Any custom responses like a product detail page or CMS in the Store API can't be used with this feature, as the Store API needs the whole entity to generate the response. If you need a small response, we recommend to use the [includes](./search-criteria.md#includes-apialias) feature of the Search API.
```
sushmangupta, line 126:
```suggestion
The current limitation of the `partial data loading` is that it only works on the Entity level. Any custom responses like a product detail page or CMS in the Store API can't be used with this feature, as the Store API needs the whole entity to generate the response. If you need a small response, we recommend to use the [includes](./search-criteria.md#includes-apialias) feature of the Search API.
```

guides/integrations-api/general-concepts/partial-data-loading.md
sushmangupta, line 109: what is meant by "the depending on fields" here ?

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+  ],
+  "aggregations": []
+}
+```
+
+Fields can reference also fields of associations like in this example the assigned salesChannel names of the currency. The API adds the necessary associations automatically for you.
+
+```http
+POST /api/search/currency
+Authorization: Bearer .....
```
sushmangupta, line 56:
```suggestion
Authorization: Bearer YOUR_ACCESS_TOKEN
```

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Partial Data Loading vs Includes
+
+Partial Data Loading is different from the [includes](./search-criteria.md#includes-apialias) feature. While includes works as post output processing, so the complete entity or data is loaded in the backend side and then filtered, partial data loading works already on Database level. This means that the database only loads the requested fields and not the whole entity.
+
+## Usage
+
+```http
+POST /api/search/currency
+Authorization: Bearer .....
```
sushmangupta, line 20:
```suggestion
Authorization: Bearer YOUR_ACCESS_TOKEN
```

sushmangupta (changes requested):

guides/integrations-api/general-concepts/partial-data-loading.md
shyim: like `isSystemDefault` is a runtime field and will be generated when the id is there. So when `id` is in fields, automatically `isSystemDefault` will also occur

guides/integrations-api/general-concepts/partial-data-loading.md
shyim: similar to here https://github.com/shopware/docs/pull/1248#discussion_r1464607562

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+## Default fields
+
+Some fields are always loaded like the `id` or join relevant fields like foreign keys, these are necessary for the API to work correctly and can't be removed.
+
+## Runtime fields
+
+Some fields in the API are generated at runtime like `isSystemDefault` of the currency. These fields are loaded by default when the depending on data is available, otherwise they can be referenced in the `fields` parameter to force the API to load them.
```
sushmangupta, line 107:
```suggestion
Some fields in the API are generated at runtime like `isSystemDefault` of the currency. These fields are loaded by default when the referenced data is available, otherwise they can be requested in the `fields` parameter to force the API to load them.
```

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+
+For custom entity definitions with runtime flag, the depending on fields needs to be specified inside the constructor. See an example from the core:
```
sushmangupta, line 109:
```suggestion
For custom entity definitions with runtime flag, the referenced fields need to be specified inside the constructor. See an example from the core:
```

guides/integrations-api/general-concepts/partial-data-loading.md
```diff
@@ -0,0 +1,126 @@
...
+
+        // When this field is requested, we need the data of path field to generate the url
+        (new StringField('url', 'url'))->addFlags(new ApiAware(), new Runtime(['path'])),
+    ]);
+}
+```
+
+## Limitations
+
+The current limitation of the `partial data loading` is that it only works on the Entity level. Any custom responses like a product detail page or CMS in the Store API can't be used with this feature, as the Store API needs the whole entity to generate the response. If you need a small response, we recommend to use the [includes](./search-criteria.md#includes-apialias) feature of the Search API.
```
sushmangupta, line 126:
```suggestion
The current limitation of the `partial data loading` is that it only works on the Entity level. Any custom responses like a product detail page or CMS in the Store API can't be used with this feature, as the Store API needs the whole entity to generate the response. If you need a small response, we recommend using the [includes](./search-criteria.md#includes-apialias) feature of the Search API.
```

sushmangupta (changes requested):

sushmangupta (approved):
Thanks for the update! :)

PR #1258, opened 2024-01-23
Fix number of remotes + default paas remote name

despokd:
1. the default remote name of PaaS for me was `shopware` instead of `platform`
2. Number of remotes is 2 (`origin` and `shopware`) instead of 3

My resulting setup after following the docs:
image


Isengo1989 (approved):

Isengo1989:
Thanks for the PRs @despokd :tada:

PR #1284, opened 2024-02-07
Test reviewdog

Isengo1989:


Isengo1989:
@sushmangupta @bojanrajh  as discussed in the last daily, it would be good to find the rule that was triggered on antipattern, but no luck here. If you see a PR with a spelling issue on Reviewdog, please let me know. Atm it is not reproducible and therefor, I do not know what to exclude as a rule.

PR #1286, opened 2024-02-09
Add minimum mysql 8.0 version

AydinHassan:
We did some testing today and found that MySQL 8.0.17 is the minimum version where our tests pass. So we should probably state that as the min version.


AydinHassan:
@Isengo1989 I think we can merge this now :)

Isengo1989 (approved):
Thx for the PR :tada:

PR #1300, opened 2024-02-19
don't overwrite host header in varnish vcl

jonashrem:
Please don't overwrite the host header in the varnish vcl. In most cases you can just forward the host header from the client.

In other cases, this is nothing specific to "soft purge" it isn't done in the other two vcls and a rare edge case. 

Copying this line, will actually break stores through.


shyim:
discovered that too yesterday 😅

PR #1305, opened 2024-02-23
Update add-custom-controller.md

thijsMemoICT:
Changed type for routes.xml as mentioned in https://github.com/shopware/shopware/issues/3566


Isengo1989:
We should also double-check the other mentions, for example:

> Furthermore, we have to define our `Route` with `defaults` and `_routeScope` via annotation, it is used to define which domain a route is part of and **needs to be set for every route**. In our case the scope is `storefront`.

Should be renamed to **via attribute**

ssltg (approved):

PR #1310, opened 2024-02-27
Fix default

Isengo1989:
https://shopwarecommunity.slack.com/archives/C01JH7CGT7V/p1709035100853629


sushmangupta (approved):
Thanks for the update! :)

PR #1315, opened 2024-02-29
Formatting fix and question

amenk:
Hopefully fixing this:

image

Bonus question:

> We always test with the [actual SW6 version](https://www.shopware.com/de/download/#shopware-6). So set it to the
> actual SW6 version e.g., shopware/testenv:6.4.3. Always test with the app`s highest supported Shopware version.

What does actual mean here? Is this bad translation from "aktuelle" ? Or does it mean the one selected? Should be clarified


Isengo1989 (approved):
Thanks for the PR. Tested locally and it does fix it :+1: 

In regards of your bonus question, we will get back to you

PR #1324, opened 2024-03-08
Expand rule docs

jasperP98:
The current documentation overlooks adding a new group, which would be quite useful to include in the default example. Additionally, it doesn't highlight the efficiency gain achievable by utilizing the \Shopware\Core\Framework\Rule\Rule::getConfig function. This feature seems really handy, particularly for simpler rules, as it could save a significant amount of time and work.

### DIFF
--- a/guides/plugins/plugins/framework/rule/add-custom-rules.md
+++ b/guides/plugins/plugins/framework/rule/add-custom-rules.md
@@ -119,7 +119,7 @@ $customer = $scope->getSalesChannelContext()->getCustomer();
 $loggedIn = $customer !== null;
 ```
 
-It is possible to add config to our rule. This makes it possible to skip the [Custom rule component](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-component) and the [Custom rule Administration template](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-administration-template) parts.
+It is possible to add config to our rule. This makes it possible to skip the [Custom rule component](#custom-rule-component) and the [Custom rule Administration template](#custom-rule-administration-template) parts.
 
 ```php
     public function getConfig(): RuleConfig
@@ -128,7 +128,7 @@ It is possible to add config to our rule. This makes it possible to skip the [Cu
     }
 ```
 
-when [Showing rule in the Administration](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#showing-rule-in-the-administration) we would not use a custom component but we would render the `sw-condition-generic` component.
+when [Showing rule in the Administration](#showing-rule-in-the-administration) we would not use a custom component but we would render the `sw-condition-generic` component.
 
 
 ### Active rules
--- a/guides/plugins/plugins/framework/rule/add-custom-rules.md
+++ b/guides/plugins/plugins/framework/rule/add-custom-rules.md
@@ -119,6 +119,18 @@ $customer = $scope->getSalesChannelContext()->getCustomer();
 $loggedIn = $customer !== null;
 ```
 
+It is possible to add config to our rule. This makes it possible to skip the [Custom rule component](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-component) and the [Custom rule Administration template](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-administration-template) parts.
+
+```php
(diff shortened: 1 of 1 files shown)


Isengo1989:
@jasperP98 thx for the PR :+1: 

I think it would make sense to add the group info in a separate section/subsection. Otherwise a lot of extensions will end up with groups and that might be a pain for some customer that have a lot of plugins. Might also be an option to highlight the group with https://vitepress.dev/guide/markdown#line-highlighting-in-code-blocks

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -119,6 +119,18 @@ $customer = $scope->getSalesChannelContext()->getCustomer();
 $loggedIn = $customer !== null;
 ```
 
+It is possible to add config to our rule. This makes it possible to skip the [Custom rule component](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-component) and the [Custom rule Administration template](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#custom-rule-administration-template) parts.
```
bojanrajh, line 122:
```suggestion
It is possible to add config to our rule. This makes it possible to skip the [Custom rule component](#custom-rule-component) and the [Custom rule Administration template](#custom-rule-administration-template) parts.
```

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -119,6 +119,18 @@ $customer = $scope->getSalesChannelContext()->getCustomer();
...
+
+```php
+    public function getConfig(): RuleConfig
+    {
+        return (new RuleConfig())->booleanField('isFirstMondayOfTheMonth');
+    }
+```
+
+when [Showing rule in the Administration](https://developer.shopware.com/docs/guides/plugins/plugins/framework/rule/add-custom-rules.html#showing-rule-in-the-administration) we would not use a custom component but we would render the `sw-condition-generic` component.
```
bojanrajh, line 131:
```suggestion
when [Showing rule in the Administration](#showing-rule-in-the-administration) we would not use a custom component but we would render the `sw-condition-generic` component.
```

bojanrajh (changes requested):

jasperP98:
@Isengo1989 I have moved the group docs to a different subsection and i used the vitepress docs to highlight the group.

guides/plugins/plugins/framework/rule/add-custom-rules.md
```diff
@@ -161,6 +173,39 @@ import './decorator/rule-condition-service-decoration';
 It may be possible that rules, with your newly created condition, aren't selectable in some places inside the Administration — for example, inside the promotion module. That is because rules are "context-aware". To learn more about that feature [click here](#context-awareness)
 :::
 
+#### Creating a new group in the administration
+The rule will now be added to the list of rules in the admin. It might be useful to create a new group for your rules. We can create a new group by using the `upsertGroup` method of the [RuleConditionService](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Administration/Resources/app/administration/src/app/service/rule-condition.service.js).
```
Isengo1989, line 177: @jasperP98 any reason you used the 6.3 docs here? It currently is flagged as 6.5

The files were changes from js to ts by then. Function is still available tho.

https://github.com/shopware/shopware/blob/v6.5.8.7/src/Administration/Resources/app/administration/src/app/service/rule-condition.service.ts

Same goes for 6.6 so I am currently thinking to go with 6.6

jasperP98: @Isengo1989 no specific reason actually, i will change that ASAP

Isengo1989: @jasperP98 thx. Could you please also use // [!code focus] instead of highlight? I double checked locally and highlight is not supported yet. Will have to create an internal PR for that first.

Isengo1989: image

Example

jasperP98: @Isengo1989 I updated the PR. It looks like focus does not work either.

Isengo1989: @jasperP98 thx, for me it worked (see above screenshot). Should be merged next week, the colleague who tagged it 6.5 is currently on vacation. Just making sure he had no direct issue with it being merged to 6.6 @bojanrajh fyi

jasperP98: @Isengo1989 okay great thanksi

PR #1325, opened 2024-03-11
NEXT-26247 - update component registration for asynchronous components

jleifeld:
The current documentation is describing how to register custom components in a synchronous way. The preferred way are lazy loaded components which doesn't affect the load time of the administration that much.

I updated the part of the documentation so that it explains both ways and why the asynchronous way is the preferred one.


sushmangupta (approved):
Thanks for the update! :)

PR #1334, opened 2024-03-24
Fix varnish docs to reflect changes in version 6.6

schneider-felix:
In Shopware version 6.6 the config key changed from storefront to shopware. I fixed the alert to reflect those changes.

### DIFF
--- a/guides/hosting/infrastructure/reverse-http-cache.md
+++ b/guides/hosting/infrastructure/reverse-http-cache.md
@@ -21,7 +21,9 @@ In this guide, we will use Varnish as an example for HTTP cache.
 ### The example Setup with Varnish
 
 ::: warning
-This setup is compatible from Shopware version 6.4.
+This setup is compatible from Shopware version 6.6.
+
+The configuration key changed in version 6.6. Please make sure to use the right docs for your version of Shopware.
 :::
 
 ![](../../../assets/hosting-infrastructure-reverseHttpCache.svg)


JoshuaBehrens (approved):

Isengo1989 (changes requested):
@schneider-felix not sure if I follow here. The varnish feature is available since 6.4 - replacing it with 6.6 would be incorrect.

Would recommend to add a warning section to the actual config, not above to have it in clear view

image

schneider-felix:
@Isengo1989 first of all, thanks for your feedback. Here is why I did it that way:

- In its current state the warning says `This setup is compatible from Shopware version 6.4.` which it is definitely not.

In version 6.6 the configuration changed from:
```yaml
storefront:
    http_cache:
```
to this:
```yaml
shopware:
    http_cache:
```
- This warning is not specific to this code block. There are multiple code blocks on this page, all using this configuration key.

You might miss this warning entirely when you navigate to this page using the search, because it can jump to a specific heading in the document. This is an issue because it is very common to just quickly look something up without changing to the correct version. We could just add a warning alert to every code block in this document if that's okay with you.

I'm looking forward to your feedback on this.

We just encountered this issue during the Shopware Community Unconference and this change is really easy to miss the way it is documented right now.

Isengo1989:
> @Isengo1989 first of all, thanks for your feedback. Here is why I did it that way:
> 
> * In its current state the warning says `This setup is compatible from Shopware version 6.4.` which it is definitely not.
> 
> In version 6.6 the configuration changed from:
> 
> ```yaml
> storefront:
>     http_cache:
> ```
> 
> to this:
> 
> ```yaml
> shopware:
>     http_cache:
> ```
> 
> * This warning is not specific to this code block. There are multiple code blocks on this page, all using this configuration key.
> 
> You might miss this warning entirely when you navigate to this page using the search, because it can jump to a specific heading in the document. This is an issue because it is very common to just quickly look something up without changing to the correct version. We could just add a warning alert to every code block in this document if that's okay with you.
> 
> I'm looking forward to your feedback on this.
> 
> We just encountered this issue during the Shopware Community Unconference and this change is really easy to miss the way it is documented right now.

Thank you for the PR as well and notifying us :+1: 

This setup refers to varnish, not the config. Referring to it here as "This setup is compatible from Shopware version 6.6." would rob developers from the option to test the feature on 6.4 and 6.5 when browsing the default docs.

The jumping / search would also diminish your initial solution in the top warning I suppose. 

Too many warnings can look and feel a little bit overwhelming. So I would stick with the initial thought of the general config at the top. An additional comment in the following code blocks can be possible as well, if you feel like it would help other devs.

Isengo1989:
@schneider-felix ping

We did discuss it in the team and should go with the comment per code block as mentioned above. Will you add the changes? Otherwise, I will add it by the end of the week.

Isengo1989:
Since there was no feedback given I will close the PR and create a new one (commit suggestions are not possible in the extend of adding multiple comments).

New PR https://github.com/shopware/docs/pull/1350

PR #1353, opened 2024-04-13
Create update.md

Schrank:
Start a page about infos for updates

### DIFF
--- a/guides/installation/update.md
+++ b/guides/installation/update.md
@@ -1,0 +1,3 @@
+# Update
+
+More infos about breaking changes can be found on github in our [UPGRADE overview](https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.6.md).


Isengo1989:
@Schrank there is a basic update guide in the user docs: https://docs.shopware.com/de/shopware-6-de/update-guides/update-guide-shopware-66?category=shopware-6-de/update-guides

 However, as for the developer docs we recommend the upgrade guide in the repository
 
 https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.6.md

Isengo1989:
@Schrank we discussed the topic internally and come up with a solution to add the UPGRADE.md files from shopware/shopware in the future. This will not be static, but pulled on every change from shopware/shopware, so closed for now. Thx for sharing.

PR #1354, opened 2024-04-15
Adjust code examples

Isengo1989:
https://shopwarecommunity.slack.com/archives/C01JH7CGT7V/p1713052793839999


jleifeld (changes requested):
I would also add that you need to create a computed which creates the `assetFilter`. Example: https://github.com/shopware/docs/pull/1366

jleifeld (approved):

PR #1361, opened 2024-04-18
remove twig cache changing path

shyim:
This will be deprecated in Symfony 7.1, and it makes no sense to move it

### DIFF
--- a/guides/hosting/performance/caches.md
+++ b/guides/hosting/performance/caches.md
@@ -94,14 +94,3 @@ framework:
         default_redis_provider: 'redis://host:port'
 ```
 
-## Twig cache
-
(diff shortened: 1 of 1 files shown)


Isengo1989:
@shyim is it already removed and we are running 7.1 (since 7.1 is only dev so far) ?

shyim:
we are running 7.0 and it will be deprecated in 7.1

Isengo1989:
@shyim I don't think this is used very commonly, but we will leave the PR open until the functionality is fully removed. I will add the label blocked for now.

Isengo1989 (approved):
Thank you for the PR :tada: 

Merged as soon as the version is released

PR #1374, opened 2024-04-26
fix: remove info about encrypted message bus

tinect:
There is no `encrypted.messenger.bus.shopware` since shopware 6.5. It got removed without any dedicated info :-( 

Please fix this info as well for docs at shopware 6.5 and add an info for older versions, so that no one will use it.


Isengo1989 (approved):
Thanks for PR @tinect  :tada:

PR #1376, opened 2024-05-05
Update customize-templates.md

erguenyogurtcu:
The “src” folder doesn't apply to apps, only to plugins. (see also here: https://developer.shopware.com/docs/guides/plugins/apps/storefront/apps-as-themes.html#migrating-existing-themes)


Isengo1989 (approved):
Thanks for your PR @erguenyogurtcu :tada:

PR #1379, opened 2024-05-08
Update template.md

JensBrunken:


guides/installation/template.md
```diff
@@ -267,7 +267,7 @@ composer require --dev profiler
 
 ### framework:demo-data is missing faker classes
 
-Prior to Shopware 6.4.17.0, you have to install some packages to get `framework:demo-data` command working:
+Prior to Shopware 6.4.17.0, you have to install some packages to get `bin/console framework:demodata APP_ENV=prod` command working:
```
bojanrajh, line 270: @Isengo1989 Should we also update [user docs](https://docs.shopware.com/en/shopware-6-en/update-guides/updating-shopware)? Is this change necessary?

Isengo1989: Yes, indeed, without it is not executing, at least in a dev enviroment

```
www-data@4adc3826def7:~/html$ bin/console framework:demodata
Demo data command should only be used in production environment. You can provide the environment as follow `APP_ENV=prod framework:demodata`
```

BUT, it should be passed as an environment variable, and not as a arg to the command (this PR therefor can not be merged)


```
In ArgvInput.php line 193:
                                                                               
  [Symfony\Component\Console\Exception\RuntimeException]                       
  No arguments expected for "framework:demodata" command, got "APP_ENV=prod". 
```


So it would be 

**APP_ENV=prod bin/console framework:demodata**


However, the hint that it should only be executed on prod environments is valid. Simply adding it might be misleading or not wanted. I will gather some "intel" from the core team :slightly_smiling_face:

Isengo1989:
@JensBrunken thx for the PR, but we can not merge it. The hint on a dev environment is wanted and the current code change you made is failing as well -> see https://github.com/shopware/docs/pull/1379#discussion_r1609519740

The hint is given when in dev as well

image

PR #1386, opened 2024-05-22
Remove set and add link

Isengo1989:


Isengo1989:
@shyim we refer to

image

obj.hits > 0


but varnish to 

image

obj.uncacheable

Is this correct or we need tu adjust?

shyim (approved):

PR #1390, opened 2024-05-24
NEXT-31522 - Added documentation for entity attributes

OliverSkroblin:


guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+
+# Entities via attributes
+
+Since shopware v6.6.3.0 it is possible to register entities via php attributes. This is a new way to register entities in Shopware 6. This guide will show you how to do this.
+
+## Define the entity
+
+First, you need to define your entity. This is done by creating a new class and adding the `Entity` attribute to it. The attribute requires the `name` parameter, which is the name of the entity. The `name` parameter is required and must be unique.
+
+Also, you have to extend the `EntityStruct` class and define a primary key. The primary key is defined by adding the `PrimaryKey` attribute to a property. In theory, the primary key can be of any type, but it is recommended to use a `UUID`.
```
bojanrajh, line 16: `EntityStruct` is an alias of `Shopware\Core\Framework\DataAbstractionLayer\Entity`. IMO, it would make sense to de-alias it and use alias for `Shopware\Core\Framework\DataAbstractionLayer\Attribute\Entity as EntityAttribute` instead?

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+## Define the entity
+
+First, you need to define your entity. This is done by creating a new class and adding the `Entity` attribute to it. The attribute requires the `name` parameter, which is the name of the entity. The `name` parameter is required and must be unique.
+
+Also, you have to extend the `EntityStruct` class and define a primary key. The primary key is defined by adding the `PrimaryKey` attribute to a property. In theory, the primary key can be of any type, but it is recommended to use a `UUID`. 
+
+```php
+<?php
+
+namespace Shopware\Tests\Integration\Core\Framework\DataAbstractionLayer\fixture;
```
mitelg, line 21: maybe we should some Foo/Bar namespace name in the docs, instead of the "real" namespace of our tests? same in the services.xml

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+<service id="Shopware\Tests\Integration\Core\Framework\DataAbstractionLayer\fixture\AttributeEntity">
+    <tag name="shopware.entity"/>
+</service>
+```
+
+That's it. You entity is registered and you can read and write data to it over the DAL. Over the tag, shopware automatically registers an `EntityDefinition` for the entity and an `EntityRepository`.
+
+## Field Types
+
+To define more fields, you typically use the `Field` attribute. The `Field` attribute requires the `type` parameter, which is the type of the field. The type can be any of the `FieldType` constants.
```
mitelg, line 51: "typically" ? is there another way? :grin: 

To define more fields, add a new public property and add the `Field` attribute to it.

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+class AttributeEntity extends EntityStruct
+{
+    #[PrimaryKey]
+    #[Field(type: FieldType::UUID)]
+    public string $id;
+    
+    #[AutoIncrement]
+    public int $autoIncrement;
+
+    #[ForeignKey(entity: 'currency')]
```
mitelg, line 90: isn't it possible to use constants here? :thinking: 

```suggestion
    #[ForeignKey(entity: \Shopware\Core\System\Currency\CurrencyDefinition::ENTITY_NAME)]
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
+---
+nav:
+  title: Entities via attributes
+  position: 1
+
+---
+
+# Entities via attributes
+
+Since shopware v6.6.3.0 it is possible to register entities via php attributes. This is a new way to register entities in Shopware 6. This guide will show you how to do this.
```
Bird87ZA, line 10:
```suggestion
Since Shopware v6.6.3.0 it is possible to register entities via PHP attributes. This guide will show you how to do this.
```

Removed redundant text

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
Bird87ZA, line 16:
```suggestion
First, you need to define your entity. This is done by creating a new class extending `EntityStruct` and adding the `Entity` attribute to it. The attribute requires the `name` parameter, which is the name of the entity. The `name` parameter is required and must be unique.

You have to define a primary key. The primary key is defined by adding the `PrimaryKey` attribute to a property. In theory, the primary key can be of any type, but it is recommended to use a `UUID`. 
```

Linguistically makes more sense

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+
+To register the entity, you have to add this class to the DI container in the `services.xml` file. This is done by adding the `shopware.entity` tag to the service definition. 
+
+```xml
+<service id="Shopware\Tests\Integration\Core\Framework\DataAbstractionLayer\fixture\AttributeEntity">
+    <tag name="shopware.entity"/>
+</service>
+```
+
+That's it. You entity is registered and you can read and write data to it over the DAL. Over the tag, shopware automatically registers an `EntityDefinition` for the entity and an `EntityRepository`.
```
Bird87ZA, line 47:
```suggestion
That's it. You entity is registered and you can read and write data to it over the DAL. Using the tag, Shopware automatically registers an `EntityDefinition` and `EntityRepository` for the entity.
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+    #[Field(type: FieldType::INT)]
+    public ?int $int;
+
+    // ...
+}
+```
+
+All field types are defined in the [`FieldType`](https://github.com/shopware/shopware/blob/trunk/src/Core/Framework/DataAbstractionLayer/Attribute/FieldType.php) class.
+
+We also provide a list of special field-types, which implement a specific behavior. They have an own php attribute class, for example, the `AutoIncrement` or `ForeignKey` field.
```
Bird87ZA, line 77:
```suggestion
We also provide a list of special field types, which implement a specific behavior. They have their own PHP attribute class, for example the `AutoIncrement` or `ForeignKey` field. 
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+    /**
+     * @var array<string, mixed>
+     */
+    #[CustomFields]
+    public array $customFields;
+}
+
+```
+
+## Api encoding
```
Bird87ZA, line 135:
```suggestion
## API encoding
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+    public int $autoIncrement;
+
+    #[ForeignKey(entity: 'currency')]
+    public ?string $foreignKey;
+}
+```
+
+## Json fields 
+
+If you want to store json data in a field, with an own validation and serialize logic, you can use the `Serialized` attribute, and define an own serializer class:
```
Bird87ZA, line 97: JSON is an abbreviation and the standard way to write it is upprecase.

```suggestion
## JSON fields

If you want to store json data in a field, with its own validation and serialization logic, you can use the `Serialized` attribute, and define its own serializer class:
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+     */
+    #[CustomFields]
+    public array $customFields;
+}
+
+```
+
+## Api encoding
+
+Each field of an entity is by default not exposed in the API. To expose a field in the API, you have to set the `api` parameter of the `Field` attribute to `true`, or to one of the scopes you want to allow.
```
Bird87ZA, line 137:
```suggestion
By default, each field of an entity is not exposed in the API. To expose a field in the API, you must set the api parameter of the Field attribute to true or specify one of the scopes you want to allow.
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+    #[Field(type: FieldType::STRING, api: [SalesChannelApiSource::class])]
+    public string $storeOnly;
+```
+
+## Translated fields
+
+To also support shopware translations for your entity, you can set the `translated` property of the `Field` attribute to `true`. This will automatically create a `TranslatedField` for the field and
+registers a `EntityTranslationDefinition` for you.
+
+You can also define a `Translations` attribute on a property to allow loading all translations of the entity.
```
Bird87ZA, line 170:
```suggestion
To support Shopware translations for your entity, set the `translated` property of the `Field` attribute to `true`. This will automatically create a `TranslatedField` for the field and
register an `EntityTranslationDefinition` for you.

Additionally, you can define a `Translations` attribute on a property to enable loading all translations of the entity.
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+     */
+    #[Translations]
+    public array $translations;
+}
+```
+
+## Required fields
+
+Each field which is not type hinted as nullable is required by default. But you can also mark a field as required by adding the `Required` attribute to the field. This will automatically add a validation rule to the field.
+This is necessary for fields which are marked as `translated`, because the translated fields need to be nullable.
```
Bird87ZA, line 196:
```suggestion
By default, any field that is not type-hinted as nullable is required. However, you can explicitly mark a field as required by adding the `Required` attribute to it. This will automatically add a validation rule to the field.
This is necessary for fields marked as `translated`, as translated fields must be nullable.
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+
+    #[Required]
+    #[Field(type: FieldType::STRING, translated: true)]
+    public ?string $required = null;
+}
+```
+
+## Associations
+
+It is also possible to define associations between entities. You can you one of the following four association types `OneToOne`, `OneToMany`, `ManyToOne` and `ManyToMany`.
```
Bird87ZA, line 214:
```suggestion
It is also possible to define associations between entities. You can use one of the following four association types: `OneToOne`, `OneToMany`, `ManyToOne` or `ManyToMany`.    
```

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+
+All the associations are defined as a nullable array property. The key of the array is the id of the associated entity. The value is the associated entity itself.
+
+You can also typehint to many associations with the `EntityCollection` class.
+
+## Getter & Setter, Translations and Collections
+
+With this new pattern, we removed the need for getter and setter methods. The properties are public and can be accessed directly.
+Also, you don't have to define any `EntityTranslationDefinition` or `EntityCollection` anymore, which reduces the boilerplate code. 
+
```
Bird87ZA, line 260: remove empty line

Bird87ZA (commented):
Very cool. Super job. 🎉

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
bojanrajh: Maybe use some expected namespace for Plugins?

Haberkamp (dismissed):

bojanrajh (approved):

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -0,0 +1,411 @@
...
+    public int $autoIncrement;
+
+    #[ForeignKey(entity: 'currency')]
+    public ?string $foreignKey;
+}
+```
+
+## JSON fields
+
+If you want to store json data in a field, with its own validation and serialization logic, you can use the `Serialized` attribute, and define its own serializer class:
```
sushmangupta, line 97:
```suggestion
If you want to store JSON data in a field with its own validation and serialization logic, you can use the `Serialized` attribute and define its own serializer class:
```

sushmangupta (changes requested):

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
OliverSkroblin: yes it's possible, but it makes it harder to read. I would not recommend this. we have a validator and if you failed in the name, it throws an exception.

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
OliverSkroblin: yes here are other ways, other field classes.

sushmangupta (approved):

PR #1391, opened 2024-05-29
feat: add info about making extension manager readonly

shyim:


guides/hosting/installation-updates/extension-managment.md
```diff
@@ -93,3 +93,17 @@ And run the below command to re-generate the class loader.
...
+
+To disable the installation of extensions in the Administration, you can set the following configuration in your `config/packages/z-shopware.yaml` file:
+
+```yaml
+shopware:
+    deployment:
+        runtime_extension_management: false
+```
+
+and clear once the cache. After that, the Extension Manager in the Administration will be read-only, so you can access only the extension configuration also the First Run Wizard will not download extensions like PayPal or the Shopware Store.
```
nickygerritsen, line 109: Some nitpicking about wording:
```suggestion
and clear the cache once. After that, the Extension Manager in the Administration will be read-only, so you can only access the extension configuration. Also the First Run Wizard will not download extensions like PayPal or the Shopware Store.
```

sushmangupta (changes requested):

guides/hosting/installation-updates/extension-managment.md
```diff
@@ -93,3 +93,17 @@ And run the below command to re-generate the class loader.
 ```bash
 composer dump-autoload
 ```
+
+## Making the Extension Manager in Administration read-only
```
sushmangupta, line 97:
```suggestion
## Configuring Extension Manager to read-only in Admin
```

PR #1405, opened 2024-06-10
feat(nginx): block php files inside user uploaded directories

shyim:


resources/references/config-reference/server/nginx.md
```diff
@@ -23,6 +23,11 @@ server {
     location ~ ^/shopware-installer\.phar\.php/.+\.(?:css|js|png|svg|woff)$ {
      try_files $uri /shopware-installer.phar.php$is_args$args;
     }
+
+    # Block .php .git .htaccess .env files in public folders
+    location ~ ^/(media|thumbnail|theme|bundles|sitemap).*\.(php|git|htaccess|env)$ {
```
bojanrajh, line 28: Usually, all files and dirs starting with dot (.) should be denied, except for `.well-known`. This also covers git, htaccess and env, but we need a different rule for denying .php files.

```
# Allow access to .well-known for letsencrypt
location ^~ /.well-known/ {
    allow all;
}

# Deny access to files and dirs starting with . (dot)
location ~ /\. {
    deny all;
}
```

shyim: @linushstge

linushstge: I agree to @bojanrajh, but let's also ensure to block .php files inside public directories to prevent any unwanted execution of uploaded files from third party plugins 

```
# Deny access to . (dot) files
location ~ /\. {
    deny all;
}

# Deny access to .php files in public directories
location ~ ^/(media|thumbnail|theme|bundles|sitemap).*\.php$ {
    deny all;
}
```

bojanrajh (approved):

PR #1412, opened 2024-06-18
docs: add missing PHP import in storefront controller example

cngJo:
When copy/pasting the example, the `#[Route()]` use statement is missing 😉


sushmangupta (approved):
Thanks for the update! :)

PR #1423, opened 2024-06-27
v6.7.0.0-dev

bojanrajh:
This PR will be merged once `v6.7` is released and `v6.6` is branched out of `main`.


Isengo1989:
@keulinho, when is the best time to run this script and update the references? Do we have other references that we need to refresh?




https://github.com/shopware/docs/blob/main/resources/references/app-reference/script-reference/script-hooks-reference.md#account-payment-method-page-loaded

guides/plugins/plugins/storefront/customize-templates.md
```diff
@@ -65,9 +79,11 @@ To override it now, just add the very same block into your custom file and repla
 {% endblock %}
 ```
 
-If you wanted to append your text to the logo instead of replacing it, you could add a line like this to your override: <code v-pre>{{ parent() }}</code>
+If you wanted to append your text to the logo instead of replacing it, you could add a line like this to your override: `{{ parent() }}`
```
bojanrajh, line 82:
```suggestion
If you wanted to append your text to the logo instead of replacing it, you could add a line like this to your override: <code v-pre>{{ parent() }}</code>
```

PR #1431, opened 2024-07-08
NEXT-37105 - new payment handling

lernhart:


guides/plugins/plugins/checkout/payment/add-payment-plugin.md
```diff
@@ -7,6 +7,14 @@ nav:
 
 # Add Payment Plugin
 
+::: warning
+With Shopware 6.6.5.0, the payment handling was refactored.
+Most of this documentation is deprecated and obsolete with 6.7.0.0.
+
+The new payment handling is done via a single `AbstractPaymentHandler`.
```
Isengo1989, line 14: @lernhart the current branch next-6.7 is in progress at the moment and typically will only be merged when the release is out. Was there a different approach discussed with @bojanrajh ? Or is all of it already available in 6.6.5.0 ?

lernhart: It is available with 6.6.5.0 and deprecated. I discussed this with the Docs team. We should add a 6.7 (upcoming) version to the selector in the Docs, so developers can see the upcoming changes. 

If you have a look at other documentations, symfony for example, you will see that they have a similar approach.

cyl3x (approved):

Isengo1989:
@bojanrajh fyi - I just tagged it 6.7 to merge when that branch goes live. Destination is v6.6

guides/plugins/plugins/checkout/payment/add-payment-plugin.md
bojanrajh: Confirmed, the `v6.7 (unreleased)` version will be added as soon as [this PR](https://github.com/shopware/docs/pull/1423) will have some changes.

PR #1436, opened 2024-07-16
Remove instructions about the theme building

vrobert78:
Remove the instructions about the theme building, which is no longer needed, with the last version of the template.


Isengo1989 (approved):
Thx for the PR @vrobert78 :tada:

PR #1437, opened 2024-07-16
Add the steps to deploy in less than 6 minutes

vrobert78:
We would like to propose a guide to have Shopware deployed on PaaS in less than 6 minutes.


Isengo1989 (approved):
Thx for the PR @vrobert78 :tada:

PR #1443, opened 2024-07-22
remove mysql hint

shyim:


JoshuaBehrens:
@shyim  Why though? Is it not an issue anymore? Do I need a specific version for it to not be an issue?

PR #1447, opened 2024-07-30
Update requirements to list MySQL 9.0 support

Gaitholabi:


AydinHassan (approved):

guides/installation/requirements.md
```diff
@@ -55,7 +55,7 @@ You can use these commands to check your actual environment:
 
 * MySQL
 
-  * Recommended version: 8.0
+  * Compatible versions: 8.0 and 9.0
```
shyim, line 58: MySQL 9 is an innovation release. I would add something like not recommended to use in production or even tbh not mention here 🤔

Gaitholabi: I tend to agree, maybe we could ignore listing it here, and if someone asks specifically, we could tell them that it should be fine but not recommended

sushmangupta: @Gaitholabi or @shyim , shall I close this PR or change "recommended version to compatible version" ?

Gaitholabi: @sushmangupta i'll close it for now 👍

PR #1456, opened 2024-08-13
NEXT-37362 - add documentation for healthchecks

Gaitholabi:


guides/plugins/plugins/framework/system-check/index.md
```diff
@@ -0,0 +1,5 @@
+# Overview
+
```
shyim, line 2: Whats missing is how you trigger it 🤔

Gaitholabi: @shyim fixed

guides/plugins/plugins/framework/system-check/index.md
```diff
@@ -0,0 +1,25 @@
...
+In this guide, you will learn about the system health check in Shopware. System health checks are a way to monitor the health of a system and detect failures early. 
+
+You can find the core concepts well-defined in the Concepts section of the documentation [System Checks](../../../../../concepts/framework/system-check.md)
+
+## Triggering System Checks
```
keulinho, line 7: WDYT about adding a clear example of how to incorporate pre rollout checks into deployment workflows?

just some lines of bash that executes the pre rollout checks and "rollbacks" if it does not succeed? 

But such stuff might be better suited in the hosting part of the docs 🤔 

What do you think about following structure:
1. general concepts under concepts (as you have it right now)
2. how to add a custom check under plugin (as you have it)
3. how to call and use the system checks under hosting (so basically this moves to hosting)

and then we should cross reference the docs 🤔

Gaitholabi: I think it makes sense to have it in both areas (you want to know how to run your system check as a part of development flow). 

I'll add it to the hosting and illustrate the concept of rollback/rollout 👍

guides/plugins/plugins/framework/system-check/add-custom-check.md
```diff
@@ -0,0 +1,98 @@
...
+        private readonly string $installationPath,
+        private readonly int $warningThresholdInMb
+    )
+    {
+    }
+
+    public function run(): Result
+    {
+        if ($this->adapterType !== 'local') {
+           return new Result(name: $this->name(), status: Status::SKIPPED, message: 'Disk space check is only available for local file systems.', healthy: true)
```
AydinHassan, line 59: Maybe a little description about the healthy flag might be useful, because it could be confusing when a result can have a status of error but also be healthy?

AydinHassan (changes requested):

keulinho (approved):

PR #1477, opened 2024-09-12
Adjust app docs to clearly mention for what features an app registrat…

keulinho:
…ion is needed


guides/plugins/apps/app-base-guide.md
```diff
@@ -92,7 +96,7 @@ The setup workflow is shown in the following schema. Each step will be explained
 The timeout for the requests against the app server is 5 seconds.
 :::
 
-## SDK Integration
+### SDK Integration
```
keulinho, line 99: this was intentionally: the sdk integration described here is for the registration, now it looks like more generic, but this whole part is still about registration

shyim (approved):

PR #1488, opened 2024-09-18
docs: add stoplight info, deprecate notice swagger ui

BrocksiNet:


Isengo1989 (approved):
Thx for the PR @BrocksiNet - Stoplight looking good and so much faster :partying_face:

PR #1495, opened 2024-09-24
Add security github page

Isengo1989:
image


Isengo1989:
After feedback from Lukas merged without review

Issue #1527, opened 2024-10-22
Missing images?


melroy89:
I don't see any diagrams or images on the documentation page, like: https://developer.shopware.com/docs/resources/references/adr/2021-10-01-payment-flow.html??

I believe something goes wrong. Like either the asset is removed or not generated correctly. Or invalid png: https://developer.shopware.com/assets/asynchronous-payment.BzwXwsjL.png

Isengo1989:
Thx @melroy89 for informing us! Should be fixed with the next deployment.

If you need the images right aways pls check our main repo -> https://github.com/shopware/shopware/tree/trunk/adr

PR #1529, opened 2024-10-23
Add asset check

Isengo1989:
https://github.com/shopware/docs/issues/1527


melroy89:
Thank you!

Ps. I'm also from 1989 😊

PR #1533, opened 2024-10-29
Add Structure to PaaS

sushmangupta:


sushmangupta:
Closing it for now as this will next be overlooked in the month of May. However, this ticket will be used as a reminder to rework on it https://shopware.atlassian.net/browse/DX-608. Informed Maike as well.

PR #1539, opened 2024-11-06
feat: add github actions example to deployer

shyim:
Key changes:

- The pipeline builds Shopware using shopware-cli
- Deployment Helper will be used for migrations and installations
- Added GitHub Action example


Isengo1989 (approved):
Thx for the PR @shyim :tada:

PR #1556, opened 2024-11-13
Update override-existing-javascript.md

Nico-Schaefer-2111:
- add note for overriding async JS-Plugin

### DIFF
--- a/guides/plugins/plugins/storefront/override-existing-javascript.md
+++ b/guides/plugins/plugins/storefront/override-existing-javascript.md
@@ -49,6 +49,18 @@ export default class MyCookiePermission extends PluginClass {
 }
 ```
 
+
+::: info
+This does not work, if the plugin you want to override is an async plugin. Currently you need to import a third-party plugin like below.
+:::
+
+```javascript
+import ThirdPartyPluginClass from '../../../../../../../../../vendor/store.shopware.com/thirdparty/src/Resources/app/storefront/src/plugin/third-party-js-plugin-dir/third-party-js.plugin';
+
+export default class MyCookiePermission extends ThirdPartyPluginClass {
+}
+```
+
 Now you can override the functions from the parent class.
 
 ### Always show the cookie bar


JoshuaBehrens:
@aragon999

aragon999:
Thank you, but I think this is not an ideal solution, as the path depends on how the plugin is installed, and it is not always possible to know how a plugin is installed, for example when one has a store plugin, which extends another store plugin. 

I tried to configure the one plugin as webpack path, to import it not via the relative path, but I did not manage to get it working. I will have a closer look into that when I am back from my holidays.

Nico-Schaefer-2111:
@aragon999 thanks, this kind of implementation is triggering me either. But I couldn't find another solution for this.

tobiasberge:
Hi, sorry for my late reply. I agree that this solution is not looking nice with the ugly path.

I think the problem is, that the prior example with `getPlugin('PluginName')` and then `get('class')` will not give you the class itself if it is async, because it was not downloaded yet.

Because the `PluginManager.override()` with the override-plugin containing the `get('class')` is happening BEFORE `initializePlugins()` which is downloading the plugins in the first place. For core plugins this is not a problem because everyone is likely just using the alias `src/` to directly import it statically.

A webpack alias could be an alternative to avoid the path at least in the code. But like @aragon999 already said this can be tricky to determine the actual path. And it also requires a custom webpack config which was not needed before. We could try to automatically generate an alias for each bundle inside the core webpack config so that you have always something like `@MyExtension1/something`, `@AnotherExtension2/something` available.

Or solving it on a programmatic level, that we have something like `fetchPlugin('MyAsyncPlugin')` to force download a plugin before the actual initializePlugins so you are able to get the class for overriding it. But not sure honestly about the side-effects and drawbacks right of the bat without further testing.

Isengo1989:
> Hi, sorry for my late reply. I agree that this solution is not looking nice with the ugly path.
> 
> I think the problem is, that the prior example with `getPlugin('PluginName')` and then `get('class')` will not give you the class itself if it is async, because it was not downloaded yet.
> 
> Because the `PluginManager.override()` with the override-plugin containing the `get('class')` is happening BEFORE `initializePlugins()` which is downloading the plugins in the first place. For core plugins this is not a problem because everyone is likely just using the alias `src/` to directly import it statically.
> 
> A webpack alias could be an alternative to avoid the path at least in the code. But like @aragon999 already said this can be tricky to determine the actual path. And it also requires a custom webpack config which was not needed before. We could try to automatically generate an alias for each bundle inside the core webpack config so that you have always something like `@MyExtension1/something`, `@AnotherExtension2/something` available.
> 
> Or solving it on a programmatic level, that we have something like `fetchPlugin('MyAsyncPlugin')` to force download a plugin before the actual initializePlugins so you are able to get the class for overriding it. But not sure honestly about the side-effects and drawbacks right of the bat without further testing.

Maybe it is better suited here then? -> https://github.com/shopware/shopware/discussions 

As for helping developers till then, what about a more generic info, like: "For async plugins you need to import it via an absolute path where your plugin is located" (since it could also be located in custom/plugins etc.)

Isengo1989:
Maybe there are new options with Vite? Since it will replace webpack in 6.7

I will close the PR for now. Feel free to create a new one or start a discussion for a solution that works in general.

aragon999:
> Maybe there are new options with Vite? Since it will replace webpack in 6.7

Afaik it will be only replaced in the administration and not the storefront (which this PR was aimed for).

tobiasberge:
That is correct, the Vite migration effects only the administration.

PR #1561, opened 2024-11-15
chore: Improve attribute entity article

mitelg:


guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
```diff
@@ -434,11 +589,36 @@ class ExampleEntity extends Entity
...
 
+
+<?php declare(strict_types=1);
+
+namespace Examples;
+
+use Shopware\Core\Framework\DataAbstractionLayer\EntityCollection;
+
+/**
+ * @internal
```
JoshuaBehrens, line 613: Why would one suggest others, that entity classes are internal. Internal compared to what?

JoshuaBehrens (changes requested):

guides/plugins/plugins/framework/data-handling/entities-via-attributes.md
mitelg: thanks for the hint! copy pasta 😅 Will fix it tomorrow.

PR #1589, opened 2024-12-03
Lychee/fix

Isengo1989:


products/extensions/migration-assistant/guides/extending-a-shopware-migration-profile.md
```diff
@@ -9,11 +9,11 @@ nav:
 
 ## Overview
 
-In this guide, you will see an example of how you can extend a Shopware migration profile of the [Shopware Migration Assistant](https://store.shopware.com/search?sSearch=Swag257162657297f). For this example, the Shopware 5 [SwagAdvDevBundle](https://github.com/shopwareLabs/SwagAdvDevBundle) plugin is migrated to the Shopware 6. For simplicity, only the local gateway is implemented.
```
mitelg, line 12: @Isengo1989 now the link only goes to the store, which is not helpful, if you want to have the migration assistant :thinking: I guess, you can also link to GitHub?

Isengo1989: @mitelg sry, missed that comment. So the link beforehand was also pointing to the store (but the old structure of search via SW5, I just adjusted the path to work with SW6). Feel free to add the Repo here if it makes more sense, sadly I was not involved yet in the migration assistant and can not make that decision.

PR #1591, opened 2024-12-04
WIP: First structure for engineering fundamentals

Phil23:


Isengo1989:
@Phil23 since the new major version is just ahead, we are giving a heads up to old PRs. There will probably be a lot of conflicts already, but even more when 6.7 will be main.

Should we keep the PR open or can we close it?

PR #1613, opened 2024-12-16
Add note about scheduled-task:run-single command

AydinHassan:


guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
```diff
@@ -117,6 +117,12 @@ In order to properly test your scheduled task, you first have to run the command
 
 Now you still need to run the command `bin/console messenger:consume` to actually execute the dispatched messages. Make sure, that the `status` of your scheduled task is set to `scheduled` in the `scheduled_task` table, otherwise it won't be executed. This is not necessary, when you're using the admin worker.
 
+## Executing the scheduled task for testing
+
+Due to the nature of how the scheduled tasks are executed (using the message queue) it can be rather cubersome to test them, especially if your scheduled task does not run very often. Using the `bin/console scheduled-task:run-single` command, you can instantly execute your scheduled task, using its name.
```
bojanrajh, line 122:
```suggestion
Due to the nature of how the scheduled tasks are executed (using the message queue) it can be rather cumbersome to test them, especially if your scheduled task does not run very often. Using the `bin/console scheduled-task:run-single` command, you can instantly execute your scheduled task, using its name.
```

Isengo1989:
@AydinHassan are you fine with re-using this section instead?

https://developer.shopware.com/docs/guides/hosting/infrastructure/scheduled-task.html#debugging-scheduled-tasks

We try to avoid duplicate examples, and this is already described in the scheduled-task article.

AydinHassan:
oops, I did not see that. i guess this is a duplicate then :)

PR #1615, opened 2024-12-16
Add info for messenger stop

Isengo1989:
@shyim not sure if we need to add this per default (might break if no consumers used?) and where to place it best.


Isengo1989:
> Let me implement this in deployment helper, so it's fixed then generic

Maybe also take a look here -> https://shopwarecommunity.slack.com/archives/C057VV7MQKF/p1734341017910729

"the issue we are facing is that we want to stop the scheduled-task:run worker before deploying and start it again after. If it keeps running during the deploy, it will crash if that deploy introduces a new task"

shyim:
I think for scheduled-tasks the go-to is to run it as a crontab job. `bin/console scheduled-task:run --no-wait` and don't do any long running tasks

Isengo1989:
Closed due to -> https://github.com/shopware/deployment-helper/pull/40/files

PR #1622, opened 2024-12-19
Update index.md

mboelt:
Please prevent the usage of !important statements in your storefront templates.

### DIFF
--- a/resources/guidelines/testing/store/quality-guidelines-plugins/index.md
+++ b/resources/guidelines/testing/store/quality-guidelines-plugins/index.md
@@ -299,6 +299,7 @@ However, you may employ `<span class="h2">`, for instance.
 ### Do not use inline-css in the storefront templates
 
 Use your own classes and let your CSS be compiled by the app.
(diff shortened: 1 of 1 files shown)


Isengo1989:
@mboelt the headline of the paragraph is "Do not use inline CSS" and the new info states not to use "!important" in templates.

Isn't this obsolete then or is it meant for the scss files not to be used in?

mboelt:
Good point. Maybe it is clearer with this one:
Please prevent the usage of !important statements in your storefront CSS.

Isengo1989:
> Good point. Maybe it is clearer with this one: Please prevent the usage of !important statements in your storefront CSS.

I would suggest to put it under an own heading  or make the headline more generic, otherwise this is a bit confusing (headline states inline)

Also noticed that app and plugin is switched in the two articles

image

I can update it if you want

mboelt:
Please feel free to do so. I appreciate.

Isengo1989:
https://github.com/shopware/docs/pull/1638/files

PR #1624, opened 2024-12-19
docs: add performance optimization tips for class loading in PHP

shyim:


guides/hosting/performance/performance-tweaks.md
```diff
@@ -308,3 +308,29 @@ we are using `moderate` everywhere. That means **a user must interact** with a l
...
++        "classmap-authoritative": true,
++        "apcu-autoloader": true,
+        "sort-packages": true
+},
+```
+
+To completely reduce the class loading at runtime, you can enable `opcache.preload` by setting it to `<project-root>/var/cache/opcache-preload.php` and `opcache.preload_user` to the user running the PHP process. This will preload all classes into the opcache on PHP-FPM startup and reduce the class loading at runtime.
+
+::: warning
+When using `opcache.preload`, you need to **restart** the PHP-FPM after each modification to reload the preloaded classes.
```
Isengo1989, line 335: @shyim modification means any changes in the required plugins, or other code, right?

Maybe we headline each step to be more precise that these can all be done seperatly

shyim: yes any php modification requires a complete php-fpm restart then

Isengo1989 (approved):
Thx for the PR @shyim :tada:

PR #1625, opened 2024-12-19
docs: adjust contribution guidelines after move to GitHub

keulinho:
Importing PRs is not necessary anymore 🎉 
But the pipeline is now actually meaningful on GH, so contributors should make sure that it suceeds


patzick (approved):
Good point!

PR #1627, opened 2024-12-23
Add another override example to devenv docs

m-waes:


Isengo1989:
@m-waes can you please fix the spellcheck and markdown issues (see Readme)

`make fix`

and 

https://github.com/shopware/docs?tab=readme-ov-file#spellcheck

PR #1649, opened 2025-01-14
Fix broken links on Mixins Documentation

zorrobert:
I fixed the broken links on https://developer.shopware.com/docs/resources/references/administration-reference/mixins.html and some of the related pages. 

We still need a description for the `generic-condition` and `user-settings` mixins, which were not listed previously https://developer.shopware.com/docs/resources/references/administration-reference/mixins.html.

### DIFF
--- a/resources/references/administration-reference/mixins.md
+++ b/resources/references/administration-reference/mixins.md
@@ -9,20 +9,22 @@ nav:
 
 This is an overview of all the mixins provided by the Shopware 6 Administration. Mixins in the Shopware 6 Administration are essentially the same in default Vue. They behave generally the same as they do in Vue normally, differing only in the registration and the way mixins are included in a component. Learn more about them in the official [Vue documentation](https://vuejs.org/v2/guide/mixins.html).
 
-Also take a look at [how to use them in your plugin](../../../guides/plugins/plugins/administration/using-mixins.md) and [how to register your own mixin](../../../guides/plugins/plugins/administration/add-mixins.md).
+Also take a look at [how to use them in your plugin](../../../guides/plugins/plugins/administration/mixins-directives/using-mixins.md) and [how to register your own mixin](../../../guides/plugins/plugins/administration/mixins-directives/add-mixins.md).
 
 ## Overview of all the mixins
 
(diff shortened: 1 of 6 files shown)


resources/references/administration-reference/mixins.md
```diff
@@ -9,20 +9,22 @@ nav:
...
-| `sw-form-field` | This mixin is used to provide common functionality between form fields | [link](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Administration/Resources/app/administration/src/app/mixin/form-field.mixin.js) |
-| `sw-inline-snippet` | Makes it possible to use snippets inline | [link](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Administration/Resources/app/administration/src/app/mixin/sw-inline-snippet.mixin.js) |
-| `validation` | Is used to validate inputs in various form fields | [link](https://github.com/shopware/shopware/blob/v6.3.4.1/src/Administration/Resources/app/administration/src/app/mixin/validation.mixin.js) |
+| `discard-detail-page-changes` | Mixin which resets entity changes on page leave or if the id of the entity changes. This also affects changes in associations of the entity | [link](https://github.com/shopware/shopware/blob/v6.6.9.0/src/Administration/Resources/app/administration/src/app/mixin/discard-detail-page-changes.mixin.ts) |
```
bojanrajh, line 18: Should we use `trunk` instead of versioned links?

Isengo1989: Agree, can you use 

https://github.com/shopware/shopware/blob/trunk/src/Administration/Resources/app/administration/src/app/mixin/rule-container.mixin.ts

instead? @zorrobert 

I was just going to fix those admin issues, there are a couple more due to a recent update

Isengo1989:
I will merge this and fix the rest in an own PR since it is an external fork. As well as the trunk link.

resources/references/administration-reference/mixins.md
zorrobert: I wasn't sure about using trunk links, so I just changed them to the latest version. I'll use trunk next time :)

PR #1651, opened 2025-01-15
Add hint for the plugin name

TrayserCassa:
I added a hint for the plugin list command and which name to use.
I'm not so happy with the name of the command `./bin/console` I guess it's just shopware cli? If you have a better idea, please let me know.


shyim (approved):

Isengo1989 (approved):
Thanks for the PR @TrayserCassa :tada: 

I added a code example to be even more precise on what to use.

PR #1667, opened 2025-01-22
Update add-scheduled-task.md

AndreasMorasch:
Removed non existing method supposedly prescribed by the ScheduledTaskHandler Interface. This is probably a remain of an older shopware version.

### DIFF
--- a/guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
+++ b/guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
@@ -101,7 +101,7 @@ class ExampleTaskHandler extends ScheduledTaskHandler
 }
 ```
 
-The respective task handler, `ExampleTaskHandler` as defined previously in your `services.xml`, will be annotated with `AsMessageHandler` handeling the `ExampleTask` class. In addition the `ScheduledTaskHandler` has to extend from the class `Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskHandler`. This also comes with one method, that you need to implement first:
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
```diff
@@ -101,9 +101,8 @@ class ExampleTaskHandler extends ScheduledTaskHandler
 }
 ```
 
-The respective task handler, `ExampleTaskHandler` as defined previously in your `services.xml`, has to extend from the class `Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskHandler`. This also comes with two methods, that you need to implement first:
+The respective task handler, `ExampleTaskHandler` as defined previously in your `services.xml`, will be annotated with `AsMessageHandler` handeling the `ExampleTask` class. In addition the `ScheduledTaskHandler` has to extend from the class `Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskHandler`. This also comes with one method, that you need to implement first:
```
Isengo1989, line 104:
```suggestion
The task handler, `ExampleTaskHandler` as defined previously in your `services.xml`, will be annotated with `AsMessageHandler` handling the `ExampleTask` class. In addition, the `ScheduledTaskHandler` has to extend from the class `Shopware\Core\Framework\MessageQueue\ScheduledTask\ScheduledTaskHandler`. This also comes with one method that you need to implement first:
```

Isengo1989 (changes requested):

jozsefdamokos (approved):

Isengo1989 (approved):

PR #1677, opened 2025-01-27
doc: add upgrade information for Pinia migration

tajespasarela:
This PR adds information about the migration from Vuex to Pinia and explains how to proceed.

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -493,6 +493,7 @@ ManyToManyIdField
 ManyToOne
 ManyToOneAssociation
 ManyToOneAssociationField
+MapState
 MediaDataSelection
 MediaDataSet
 MediaFolderDataSet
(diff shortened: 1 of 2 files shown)


jleifeld (changes requested):
Looks really good. The only thing I'm thinking about is explaining the case how to implement the composables outside the store (which are then added again into a store) so that they are directly accessible even when Pinia isn't initialized yet.
This could be useful especially for internal developer

tajespasarela:
> Looks really good. The only thing I'm thinking about is explaining the case how to implement the composables outside the store (which are then added again into a store) so that they are directly accessible even when Pinia isn't initialized yet. This could be useful especially for internal developer

@jleifeld done!

jleifeld (approved):

Isengo1989 (approved):
Thx for the PR :tada: 

Adjusted (see last commit) the structure a little and removed unnecessary "Example" li items. Makes it easier to link if questions appear.

image

seggewiss (approved):

Isengo1989:
@tajespasarela should this be merged already or tagged for milestone 6.7 ?

PR #1681, opened 2025-02-03
fix: remove .shopware-extension.yml as disallowed files

shyim:
That file contains some information which are good to know for `shopware-cli project ci`


Isengo1989 (approved):
Thx for the PR :tada:

PR #1685, opened 2025-02-05
Test admin redirect on vercel stage

Isengo1989:


Isengo1989:
@bojanrajh please take a quick look.

I added all redirects manually (Copilot was lazy and did only 6 :crying_cat_face:  ) and added a small shellscript that can be used in the future to check all added links (this will probably not be the last time we have to do so)

image

bojanrajh (approved):

PR #1690, opened 2025-02-09
Add extensions as bundle section back in

JoshuaBehrens:
Calling out to @niklaswolf and @leonrustmeier that originally had the topic https://github.com/shopware/shopware-cli/issues/201 and got their additional documentation on it. But somehow this got lost in the migration of the documentation for Shopware CLI . See original content on web archive: https://web.archive.org/web/20240804020116/https://sw-cli.fos.gg/bundle-support/ 

I rephrased it a little bit and did not contain the part where a project would list its bundles in the composer.json because there is a different example using the yaml as well and that should be sufficient.


shyim (approved):

shyim:
ahh I missed that, for a complete project it's documented

Isengo1989:
Thx for the PR @JoshuaBehrens :100:

PR #1696, opened 2025-02-13
FIX: wrong key level in object-storage config in filesystem.md

GerDner:
"options:" does not exist anymore in config and was not removed here but later in the docs. This will result in an "access denied" error as soon as file system calls are made, despite valid credentials.
It seems "visibility:" was also moved one level higher as mentioned later in the docs.

### DIFF
--- a/guides/hosting/infrastructure/filesystem.md
+++ b/guides/hosting/infrastructure/filesystem.md
@@ -92,11 +92,10 @@ shopware:
         bucket: "{{AWS_BUCKET}}"
         region: "{{AWS_REGION}}"
         endpoint: "{{AWS_ENDPOINT}}"
-        options:
-          visibility: "public"
-          credentials:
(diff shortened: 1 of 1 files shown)


janbuecker:
@shyim added the visibility option long time ago. Do you know if it's still correct?

GerDner:
Just to be clear: 

https://developer.shopware.com/docs/guides/hosting/infrastructure/filesystem.html#amazon-s3 

does not work: 

https://developer.shopware.com/docs/guides/hosting/infrastructure/filesystem.html#fallback-adapter-configuration

does indeed work.

As visibility also was moved in the example above, I assumed options were dropped completely. But i did not observe issues with visibility. Only Creds

shyim (approved):

Isengo1989:
Thx for the PR @GerDner :tada:

PR #1699, opened 2025-02-20
Update add-custom-commands.md

mohammadbharmal:
There was one class missing in the "use Symfony\Component\Console\Attribute\AsCommand" class, which is generating an error like the following:

The command defined in "NameSpace\Of\MyClassCommand" cannot have an empty name.


mitelg (approved):

mitelg:
thanks for your contribution @mohammadbharmal 👍🎉💙

PR #1725, opened 2025-03-18
fix: Improve article about adding data to the storefront

mitelg:
With SW 6.7 the page template variable is no longer available in footer and header template files.

### DIFF
--- a/guides/plugins/plugins/storefront/add-data-to-storefront-page.md
+++ b/guides/plugins/plugins/storefront/add-data-to-storefront-page.md
@@ -82,9 +82,12 @@ The next thing we need to do is register our subscriber in the DI-Container and 
 
 Now that we have registered our Subscriber to the right event, we first need to fetch the additional data we need and then add it as an extension to the pagelet.
 
-Because we are in an event of a Pagelet we should not directly call the DAL to fetch the data. Instead we should check if there is a proper store-api route to fetch our data.
-If we just wanted to add specific products data we could use the ProductListRoute. But we want to fetch data that is currently not returned in a performant way with the store-api.
-The ProductListRoute could return the data but it would return way to much data for our purpose. Because of that we will add a new store-api route for our data.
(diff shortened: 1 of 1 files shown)


bojanrajh (approved):
@Isengo1989 For reference: https://shopware-ag.slack.com/archives/CBD3SQCMC/p1742234736446219

guides/plugins/plugins/storefront/add-data-to-storefront-page.md
```diff
@@ -82,9 +82,12 @@ The next thing we need to do is register our subscriber in the DI-Container and
...
 
-Because we are in an event of a Pagelet we should not directly call the DAL to fetch the data. Instead we should check if there is a proper store-api route to fetch our data.
-If we just wanted to add specific products data we could use the ProductListRoute. But we want to fetch data that is currently not returned in a performant way with the store-api.
-The ProductListRoute could return the data but it would return way to much data for our purpose. Because of that we will add a new store-api route for our data.
+Because we are in an event of a Pagelet we should not directly call the DAL to fetch the data.
+Instead, we should check if there is a proper store-api route to fetch our data.
+If we just wanted to add specific products data we could use the ProductListRoute.
+But we want to fetch data that is currently not returned in a performant way with the store-api.
+The ProductListRoute could return the data, but it would return way too much data for our purpose.
+Because of that we will add a new store-api route for our data.
```
Bird87ZA, line 90:
```suggestion
Since we are in a `Pagelet`-event, the DAL should not be called directly to fetch data. Instead, we should check whether a suitable `store-api` route exists.

If we only needed specific product data, we could use the `ProductListRoute`. However, this does not satisfy our use case. While the `ProductListRoute` could provide this data, it would return far more information than necessary.

Therefore, we will create a new `store-api` route tailored to our needs.
```

This just reads a bit better.

Bird87ZA (approved):

PR #1727, opened 2025-03-20
fix: move commercial to extensions menu

shyim:
Right now commercial is in navigation under Plugins and is there alone, I moved it under extensions


Isengo1989 (approved):
Thx for the PR @shyim :tada: 

Small note, redirects are not handled by vitepress. We will add an workflow for that soon. If you know a tool that could handle that, please reach out :)

PR #1734, opened 2025-03-27
chore: Update the articles where a StorefrontController is used

mitelg:
Using `setTwig` is no longer necessary.
References about header and footer in custom page objects are also removed. As they are no longer part of the generic page.


Isengo1989 (approved):
Thx for the PR @mitelg :100:

PR #1739, opened 2025-04-02
[FIX] Wrong parameter order in example code

hepisec:
In the example code on the page "Add caching for Store API route" the method `generateKey` was called with parameters in wrong order.


Isengo1989 (approved):
Thx for the PR @hepisec :tada:

PR #1742, opened 2025-04-04
fix: Update new document renderer structure

En0Ma1259:
The structure of the document renderer has changed a little bit.
Content need to be generated inside the DocumentRenderer

Jira: https://shopware.atlassian.net/browse/NEXT-38766

PR's:
https://github.com/shopware/shopware/pull/6470
https://github.com/shopware/shopware/pull/5834
https://github.com/shopware/shopware/pull/5975

### DIFF
--- a/guides/plugins/plugins/checkout/document/add-custom-document-type.md
+++ b/guides/plugins/plugins/checkout/document/add-custom-document-type.md
@@ -276,18 +276,19 @@ class ExampleDocumentRenderer extends AbstractDocumentRenderer
                     $config->jsonSerialize(),
                 );
 
-                // This block is for PDF or HTML files
+                // Utilize the DocumentFileRendererRegistry to generate the content, e.g. PDF or HTML
+                // This is the recommended approach
                 $doc->setTemplate(self::DEFAULT_TEMPLATE);
                 $doc->setOrder($order);
                 $doc->setContext($context);
-
                 $content = $this->fileRendererRegistry->render($doc);
-
-                // E.g. XML or CSV filetype
-                $content = '<?xml ...>...';
-                $content = 'Id;Name;...';
-
                 $doc->setContent($content);
+
+                // Alternatively you can manually set the content here, e.g. to some XML or CSV content
+                // $content = '<?xml ...>...';
+                // or
+                // $content = 'Id;Name;...';
+                // $doc->setContent($content);
 
                 $result->addSuccess($orderId, $doc);
             } catch (\Throwable $exception) {
--- a/guides/plugins/plugins/checkout/document/add-custom-document-type.md
+++ b/guides/plugins/plugins/checkout/document/add-custom-document-type.md
@@ -9,7 +9,7 @@ nav:
 
 ## Overview
 
-Using the Shopware Administration, you can easily create new documents. This guide will teach you how to achieve the same result, which is creating a new document, using your plugin.
+This guide will show you how to add a custom document type to your plugin. This includes adding a new document type to the database, creating a renderer for it, and adding a number range.
 
 ## Prerequisites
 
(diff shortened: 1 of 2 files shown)


mstegmeyer:
This should point to `next-6.7` right? Or have the branches already changed?

guides/plugins/plugins/checkout/document/add-custom-document-type.md
```diff
@@ -339,9 +330,9 @@ Here's what the function does:
 * For each order found, the function attempts to generate a document.
 * If successful, a new `RenderedDocument` object is created. This object is then added to the `RendererResult` object as a success.
 * If an error occurs, the exception is caught and the error message is added to the `RendererResult` object as an error.
-* `The RendererResult` object is returned.
+* The `RendererResult` object is returned.
 
-In this example, we are rendering a specific template, which we will have a look at in the next section.
+Depending on the file type we either get the content with `$this->fileRendererRegistry->render()` or we need to create the content on our own.
```
mstegmeyer, line 335: Maybe we can describe better, when we use the `FileRendererRegistry` and when not?

bojanrajh:
> This should point to `next-6.7` right? Or have the branches already changed?

The ticket was created for 6.6.10.2, and PRs were merged into trunk in Jan/Feb, so it's probably okay to merge into `main`.

nguyenytran (approved):

Isengo1989:
> > This should point to `next-6.7` right? Or have the branches already changed?
> 
> The ticket was created for 6.6.10.2, and PRs were merged into trunk in Jan/Feb, so it's probably okay to merge into `main`.

Also so some backport label on one of them

Isengo1989 (approved):
Thx for the PR @En0Ma1259 :tada: 

Approved from DX-side - can't asses the comment from @mstegmeyer - I am not too familiar with the Document topic.

guides/plugins/plugins/checkout/document/add-custom-document-type.md
```diff
@@ -263,37 +263,32 @@ class ExampleDocumentRenderer extends AbstractDocumentRenderer
...
+                $doc->setOrder($order);
+                $doc->setContext($context);
+
+                $content = $this->fileRendererRegistry->render($doc);
+
+                // E.g. XML or CSV filetype
+                $content = '<?xml ...>...';
+                $content = 'Id;Name;...';
+
+                $doc->setContent($content);
```
MalteJanz, line 290: In addition to the comment https://github.com/shopware/docs/pull/1742/files#r2030572416 I find this code example a bit misleading. I would suggest something like this:

```suggestion
                // Utilize the DocumentFileRendererRegistry to generate the content, e.g. PDF or HTML
                // This is the recommended approach
                $doc->setTemplate(self::DEFAULT_TEMPLATE);
                $doc->setOrder($order);
                $doc->setContext($context);
                $content = $this->fileRendererRegistry->render($doc);
                $doc->setContent($content);

                // Alternatively you can manually set the content here, e.g. to some XML or CSV content
                // $content = '<?xml ...>...';
                // or
                // $content = 'Id;Name;...';
                // $doc->setContent($content);
                
```

@mstegmeyer correct me if I got something wrong about this, I'm not that deep into the `DocumentFileRendererRegistry` yet. If you could use that all the time I think we shouldn't even mention the manual approach at all and include an example "FileRenderer" as well in this documentation.

MalteJanz (changes requested):

guides/plugins/plugins/checkout/document/add-custom-document-type.md
mstegmeyer: I'm not too deep into it either. As I understood it, for Zugferd files it was not possible to use the FileRendererRegistry, so it seems valid to me to document it better. But I agree with your example 👍

MalteJanz (approved):

mstegmeyer (approved):

PR #1743, opened 2025-04-04
docs: add caching guide

keulinho:
With 6.7 we changed some caching related things: https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.7.md#cache-rework

However most of the changes were not yet properly documented in the docs, additionally we removed the store-api caching doc as that is not supported anymore, but that was the only place where cache invalidation was described

This was also actively asked for in the community: https://shopwarecommunity.slack.com/archives/C088R4V0R9P/p1743668872745409


guides/plugins/plugins/framework/caching/index.md
```diff
@@ -0,0 +1,250 @@
...
+### HTTP-Cache
+
+#### Modifying the cache keys
+
+Every cached item has a unique key that is used to retrieve the cached data later on. For that it is important that all the relevant information that have an effect on the data that is being cached is part of the key.
+For example if the same data is cached in multiple languages or for multiple sales channels the key must contain the language and sales channel information, so that the correct data can be retrieved later on.
+Please note that for every potential value your key part can take, a new cache entry will be created. So if you have a key part that can take 10 different values, you will have 10 times the amount of cache entries for the same data.
+
+If you add customization to your projects that lead to different versions of the page being rendered, you need to make sure that the cache key is unique for each version of the page. This can be done by adding a specific part to the cache key that shopware is generating.
+You can do so by subscribing to the `HttpCacheKeyEvent` event and add your specific part to the key.
```
shyim, line 32: maybe add an
```
::: info
The event is called on any Request, make sure that you don't use expensive operations here like Database Queries and with an external reverse proxy, this needs to be done in the Configuration there also
:::
```

guides/plugins/plugins/framework/caching/index.md
```diff
@@ -0,0 +1,250 @@
...
+{
+    $this->cacheInvalidator->invalidate([
+        CachedSystemConfigLoader::CACHE_TAG
+    ]);
+}
+```
+
+### Delayed Invalidation
+
+By default, the cache invalidation happens delayed (for both http and object caches). This means that the invalidation is not instant, but rather all the tags that should be invalidated are invalidated in a regular interval.
```
shyim, line 215: makes sense to mention here that you can force it in API request level

Isengo1989: Isn´t this covered by https://github.com/shopware/docs/pull/1743/commits/66d3fb746d647f7b6dedaa212194ef32be5c9c92#diff-5d93842ffdb748b64806380c0b37a0b231f9b67f9c3ff01e60416317406b354cR238 ?

guides/plugins/plugins/framework/caching/index.md
Isengo1989: To clarify and separate the two topics:

```
:::info
The event is called on any Request; make sure that you don't use expensive operations like Database Queries.

Also, with an external reverse proxy, the cache key might be generated on the proxy and not in your application. In that case, you need to add the key part to the reverse proxy configuration.
:::
```

wdyt ?

Isengo1989 (approved):
Thx for the PR @keulinho :tada:

guides/plugins/plugins/framework/caching/index.md
keulinho: done

guides/plugins/plugins/framework/caching/index.md
keulinho: yes, i now linked from the intro directly to how force invalidate

PR #1744, opened 2025-04-09
Fix dateformat

HoelShare:


HoelShare:
Will the next-6.7 branch be rebased at some point or do I need to do the same change there as well?

bojanrajh:
Unfortunately, ADRs are synced directly from the [shopware/shopware](https://github.com/shopware/shopware/tree/trunk/adr), so these changes will get overwritten the next time when [Update ADRs](https://github.com/shopware/docs/actions/workflows/update-adrs.yml) will run.

Kind of irrelevant now, but to answer your question: `main` is still being synced into `next-6.7` regurarly, so you wouldn't need to create another PR if that would work. We'll stop the sync once the final 6.7 is out.

PR #1747, opened 2025-04-14
On Debian, use extrepo instead of curl | bash.

nwalfield:
`extrepo` is a debian package for securely adding external repositories, like deb.sury.org, to a system.  It's much more secure than running `curl | bash`.


shyim (approved):

shyim:
didnt know that. thanks!

PR #1752, opened 2025-04-15
test: sorting workflow

Isengo1989:
Sorting become a pain recently. Due to different languages / settings of "sort" this was mixed from time to time. Also it is cumbersome to always run the command.

This PR checks if .wordlist.txt was touched and sorts the file again with **LC_ALL=C** to be precise every time.


.github/workflows/wordlist.yml
```diff
@@ -0,0 +1,33 @@
...
+            git add .wordlist.txt
+            git commit -m "chore: sort .wordlist.txt"
+          fi
+
+      - name: Push changes
+        if: github.event_name == 'pull_request'
+        run: |
+          git push origin "HEAD:${{ github.head_ref }}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
bojanrajh, line 33: Probably no need to check for `pull_request` because we have `on: pull_request` trigger, and we can combine both steps.

```suggestion
      - name: Sort .wordlist.txt
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          LC_ALL=C sort .wordlist.txt -o .wordlist.txt
          if ! git diff --quiet .wordlist.txt; then
            git config user.name "github-actions[bot]"
            git config user.email "github-actions[bot]@users.noreply.github.com"
            git add .wordlist.txt
            git commit -m "chore: sort .wordlist.txt"
            git push origin "HEAD:${{ github.head_ref }}"
          fi
```

bojanrajh (approved):

PR #1768, opened 2025-05-06
fix: getRequest is not part of HttpCacheKeyEvent

tinect:
https://github.com/shopware/shopware/blob/37135663de520a3f5b585570e52b985b05173fda/src/Core/Framework/Adapter/Cache/Event/HttpCacheKeyEvent.php#L20

there is no method  `getRequest`.


mitelg (approved):

Isengo1989:
Thx for the PR @tinect :tada:

PR #1773, opened 2025-05-14
fix: Correct link to example plugins of B2B suite guides

mitelg:


Isengo1989 (approved):
Thx for the quick fix @mitelg :100:

PR #1786, opened 2025-05-26
docs: Add guide for making API requests in administration plugins

shyim:
## Summary
- Added comprehensive documentation guide for making HTTP requests to the Shopware API in administration plugins
- Updated wordlist to include "FormData" term used in the documentation

## Test plan
- [ ] Documentation renders correctly
- [ ] Code examples are syntactically correct
- [ ] Links to other documentation pages work properly
- [ ] Spell check passes with updated wordlist

🤖 Generated with [Claude Code](https://claude.ai/code)


JoshuaBehrens (approved):
I like the new examples. In the past I had to build custom downloads in the administration. Maybe we can have a suggested snippet from your side how to convert the response into a blob being downloaded. There are several possible solutions but you might have something like this in your codebase and can be used as an example. Then we have complex input and complex output

Isengo1989 (approved):
Thx for the PR @shyim :100:

PR #1794, opened 2025-06-10
docs: add section about disabling scheduled sitemap generation

shyim:
## Summary
Add documentation for the new configuration option to disable scheduled sitemap generation.

## Related PR
This documentation is for the feature implemented in shopware/shopware#10403

## Details
Added a new section in the performance tweaks guide that explains:
- Why you might want to disable the scheduled sitemap generation (blocks message queue for large catalogs)
- How to disable it using the new configuration option
- How to set up a dedicated cronjob as an alternative
- The advantages of using this approach

## Availability
This feature will be available starting with Shopware 6.7.1.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.wordlist.txt
+++ b/.wordlist.txt
@@ -1137,6 +1137,7 @@ createSnippetFixture
 createValidatorDefinition
 cron
 croninterval
+cronjob
 cryptographic
 csrf
 css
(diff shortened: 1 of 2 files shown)


Isengo1989:
> @shyim We should probably merge it into `next-6.7`?

I would also say let's only merge it into 6.7. If it is seen as a "upgrade" enabler, we can also do both.

PR #1802, opened 2025-06-16
feat: Add Organization unit documentation

thuylt:


Isengo1989:
> Can we have some insights on the technical usage. How can we identify the organization unit from the context? During the development the way seem to be mixed and changed when the employee and organization unit are added to the context. Are there any new useful hooks and events? Are there any limitations regarding the pending order process?
> 
> With my first hands-on experience it feels like it is a useful feature and it is a great counterpart to roles.

@thuylt could you take a look here :slightly_smiling_face:

phuoccx (dismissed):

JoshuaBehrens (approved):

nguyenky (approved):

PR #1831, opened 2025-07-09
docs: Recommend using ProxySQL for database cluster management

shyim:


Isengo1989:
@shyim does proxysql require any additional settings or is it just the default database connection?

shyim:
Its the default connection

PR #1834, opened 2025-07-09
Update for the app guidelines

sdenter:
Deleted the art of environment as we don't use them like this anymore 

Added the restriction of inline java script 

Changed the restritcition of own upload fields


Isengo1989:
@sdenter I suppose this also effects the plugin quality guide, right?

Isengo1989 (changes requested):
Adjust for both guidelines

PR #1844, opened 2025-07-15
Update docker.md to add Database configuration hint

socrec:
It is a bit confusing for beginner to setup the Database Configuration step with the Docker guide. So I included a hint on line 65 to make it faster for everyone to setup their shopware system.

### DIFF
--- a/guides/installation/setups/docker.md
+++ b/guides/installation/setups/docker.md
@@ -62,7 +62,10 @@ make setup
 ```
 
 This will install Shopware itself, create an admin user with username `admin` and password `shopware`.
-Hint: For the Database Configuration step, you should put the host as `database` instead of `localhost`, which is the Docker container name.
+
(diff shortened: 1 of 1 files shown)


Isengo1989:
@shyim is there maybe a way to pass that host into that setup / container? So no worries from the developer-side to choose the correct host.

shyim (approved):

Isengo1989:
> you cant redirect localhost to database. so this is required

By _pass_ I didn't mean redirect, but configure this somehow. As it is already existing somewhere (**make setup** does work without any user input).

Idea was to pre-fill this form instead. Less friction in usage.

shyim:
I think there is an issue in shopware/shopware to respect the DATABASE_URL which is set to pre-fill that

shyim:
but this change can be merged right?

PR #1848, opened 2025-07-18
Extend explanation about running `system:setup:staging` command with `--no-interaction`

jkrzefski:


jkrzefski:
If you set `--no-interaction`, you do always need `--force`. If you omit `--force`, the command will exit before doing anything.

Isengo1989:
> If you set `--no-interaction`, you do always need `--force`. If you omit `--force`, the command will exit before doing anything.

Thx, ok, let's add the param then directly after `--no-interaction` - makes it easier to understand. I can adjust the PR as well if you're busy.

guides/hosting/configurations/shopware/staging.md
```diff
@@ -61,7 +61,7 @@ After the database is imported and the configuration is done, you can activate t
 ./bin/console system:setup:staging
 ```
 
-This command will modify the database to be used in a staging environment. You can pass `--no-interaction` to the command to avoid the interactive questions.
+This command will modify the database to be used in a staging environment. You can pass `--no-interaction` to the command to avoid the interactive questions. In that case, you need to also pass `--force` to the command, because otherwise it would exit immediately.
```
Isengo1989, line 64:
```suggestion
This command will modify the database to be used in a staging environment. You can pass `--no-interaction --force` to the command to avoid the interactive questions.
```

Isengo1989 (changes requested):
Or I can add a suggestion :smile:

PR #1851, opened 2025-07-21
fix: update installation guide

shyim:


shyim:
we don't need that anymore.

Isengo1989 (approved):
Thx for the PR @shyim :+1:

PR #1872, opened 2025-07-30
Fix/lychee retries for flaky tests

Isengo1989:
The changes in this branch update the Lychee link checker workflow in .github/workflows/validate-external-links.yml:

**Added** 
--retry-wait-time 10 to wait 10 seconds between retries.
--max-retries 3 to retry failed URLs up to 3 times.
 --timeout 30 for a 30-second timeout per request.

This makes the link check more reliable for unstable or temporarily unreachable URLs.

Also the version was changed to fix v2 - no minor

### DIFF
--- a/.github/workflows/validate-external-links.yml
+++ b/.github/workflows/validate-external-links.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: lychee Link Checker
         id: lychee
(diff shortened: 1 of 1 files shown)


bojanrajh (approved):
:tada:

PR #1884, opened 2025-08-01
Tell plugin developers how to create immutable custom field sets

JoshuaBehrens:
Document a feature from Shopware 6.3

### DIFF
--- a/guides/plugins/plugins/framework/custom-field/add-custom-field.md
+++ b/guides/plugins/plugins/framework/custom-field/add-custom-field.md
@@ -277,7 +277,7 @@ The translated labels are added to both the field and the set, which are going t
 If you have several custom fields and want to order them within a specific order, you can do so with the `customFieldPosition` property.
 
 ::: info
-If you want the custom field set to be deleteable and editable in the administration, you need to set global to false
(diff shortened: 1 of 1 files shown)


JoshuaBehrens:
This is so much worth. Not even part of the 6.3 release notes ...

guides/plugins/plugins/framework/custom-field/add-custom-field.md
```diff
@@ -244,6 +244,7 @@ use \Shopware\Core\Defaults;
 $this->customFieldSetRepository->create([
     [
         'name' => 'swag_example_set',
+        'global' => true,
```
Isengo1989, line 247: @JoshuaBehrens I would say we keep one or the other. So add the global => true and remove the warning.

When adding the customField via Code, I suppose it is 99% unwanted to be deleted via the admin.

We could also change the warning to an info and say:

_"If you want the custom field set to be deleteable and editable in the administration, you need to set global to false"_

Isengo1989 (changes requested):

Isengo1989 (approved):
Good catch. Thank you for the PR @JoshuaBehrens :1st_place_medal:

PR #1887, opened 2025-08-02
feat: describe how to set mail data via subscriber

ROBJkE:
Extend 'Add Data to Mails' guide

### DIFF
--- a/guides/plugins/plugins/content/mail/add-data-to-mails.md
+++ b/guides/plugins/plugins/content/mail/add-data-to-mails.md
@@ -91,3 +91,45 @@ Here's the respective example `services.xml`:
     </services>
 </container>
 ```
+
+## Adding data via subscriber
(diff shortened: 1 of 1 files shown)


Isengo1989 (changes requested):
Thx for the PR @ROBJkE - please check the Copilot comments. As it mentions this can be misleading in terms of overriding the whole template.

Consider using smth like this instead and briefly explain the diff between the two, or just leave setTemplateData out. If you are not modifying existing data, this can be removed.

```
$templateData = $event->getTemplateData();
    
    // Add custom data
    $templateData['customValue'] = 'some data';
    
    // Set the modified template data
    $event->setTemplateData($templateData);
```

ROBJkE:
Added inline comments, removed unnecessary constructor, removed `setTemplateData()` example as it could be misleading

PR #1893, opened 2025-08-12
feat: Add sw-app-user-id header

SimonVorgers:
Related platform changes : https://github.com/shopware/shopware/pull/11777


shyim (approved):

Isengo1989:
@SimonVorgers do you know which version this will be released in? 6.7.1.3 ? It is probably best to add a short info here, when this feature was released. We can also merge it directly then.

PR #1910, opened 2025-08-25
feat: be more opinionated with shopware setup

shyim:


sneakyvv:
Hi @shyim 
I attended multiple talks you gave and you were passionate about devenv, so I was wondering why this choice is made. Can you elaborate?

shyim:
Hey,

We got a lot of feedback from newbies that it is difficult to setup a Shopware environment with things like:

- What is Dockware, who is dasistweb can I trust them
- Devenv, why do I need to adjust my entire system for setting up Shopware. What is this syntax: ...
- Why there are so many variants to setup Shopware: Local, Docker, Devenv, DDev, ....

To address this we want to focus more on one way to install Shopware and only document this one way. And why Docker? Because it's the most common tech used by the industry, any developer have used it, there are more resources available on internet to debug things. Any project you're facing to host/develop has one setup with that

JoshuaBehrens:
Will we lose the devenv setup? devenv really outperforms a lot and uses local storage. I understand the reasons tho. Every docker hosted setup I've used sucked. If you focus on pushing docker please annoy your hosting partners to follow these guidelines. Otherwise it will be yet another annoice to debug slow IOops and learning to understand, that you need a shared drive, that you need to expose with s3 because docker swarm/k8r can be weird at times. Do not expect others to understand kubernetes only by the means of existence of a documentation. kubernetes is a complexity hell that is not easy to start in the first place. Docker at all is not easy to start in the first place.

thomasmax-97:
I use laravel valet now for years, way better performance than dockware. For Database, Redis, RabbitMQ etc. i use docker containers to keep my OS clean. With local PHP setup using laravel valet its way more easy to setup and use XDebug. Performance is as mentioned way better this way since PHPStorm already use a lot of RAM.

PR #1917, opened 2025-08-27
feat: add crud element docs

shyim:


snippets/config/app/permissions.xml
```diff
@@ -4,10 +4,10 @@
         ...
     </meta>
     <permissions>
-        <read>product</read>
-        <create>product</create>
-        <update>product</update>
+        <!-- CRUD shortcut - equivalent to read, create, update, delete for product (since 6.7.3.0) -->
+        <crud>product</crud>
```
Isengo1989, line 8: Would like to add a code block instead where we show both, otherwise the usage of single operations will be lost



I will add the syntax and code later if this is :ok:

Isengo1989 (changes requested):

Isengo1989 (approved):
Thanks for the PR @shyim :100:

Isengo1989 (approved):

snippets/config/app/permissions.xml
```diff
@@ -1,16 +1,17 @@
...
-        <read>product</read>
-        <create>product</create>
-        <update>product</update>
+<manifest xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/shopware/shopware/trunk/src/Core/Framework/App/Manifest/Schema/manifest-3.0.xsd">
+  <meta>
+    ...
+  </meta>
+  <permissions>
+    <!-- CRUD shortcut - equivalent to read, create, update, delete for product (since 6.7.3.0) -->
```
mitelg, line 8: @shyim here the comment is wrong :grimacing:

Isengo1989: @mitelg thx, fixed :+1: 

https://github.com/shopware/docs/commit/50a83649c6d9642779b37aa1b0d17d0b5e12728d

PR #1929, opened 2025-09-11
Remove image proxy setup instructions from Docker guide

shyim:
We should stick to one solution for a problem :)


Isengo1989:
We should not remove the whole context here and just link to another page. I would add a small section explaining what "Image Proxy" means, especially since this also aims at beginners.

As a second note, the original content explained the Image proxy with settings without **shopware-cli**—this adds a layer of complexity. We don't have **shopware-cli** as a Prerequisite set on the article. The question is, do we want that in an easy setup?

PR #1973, opened 2025-10-20
Feat: Add section about removal of custom fields

JoshuaBehrens:
With recent feedback on one of my pull requests I think this should be part of the docs


JoshuaBehrens:
Thank you for merging :)

PR #1976, opened 2025-10-20
Update deployment-with-deployer.md

wannevancamp:
See https://github.com/deployphp/deployer/pull/4090/files

### DIFF
--- a/guides/hosting/installation-updates/deployments/deployment-with-deployer.md
+++ b/guides/hosting/installation-updates/deployments/deployment-with-deployer.md
@@ -329,6 +329,7 @@ set('shared_dirs', [
     'files',
     'var/log',
     'public/media',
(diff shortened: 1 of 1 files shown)


wannevancamp:
> Can you link to an example or PR in shopware/shopware, I couldn't find anything related to a dedicated public/plugins folder.

The documentation is not very clear about this. When using Flysystem as described [here](https://developer.shopware.com/docs/guides/plugins/plugins/framework/filesystem/filesystem.html), a folder named `public/plugins` will be created. This public folder needs to be shared across releases.

Isengo1989 (approved):
Thx for the PR @wannevancamp :tada: 

I will also add an example folder / file structure to the docs, as it is missing IMO in the current docs.

image

PR #1985, opened 2025-10-28
fix: Change nodeJS requirement and update recommendations

mitelg:
The minimum nodeJS version documented is wrong. 
See [6.7](https://github.com/shopware/shopware/blob/v6.7.3.1/src/Administration/Resources/app/administration/package.json#L204) and [6.6](https://github.com/shopware/shopware/blob/v6.6.10.8/src/Administration/Resources/app/administration/package.json#L245)
I also updated the recommendations to use newer versions.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,3 @@ indent_size = 2
 charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
-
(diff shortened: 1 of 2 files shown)


Isengo1989 (approved):
Thx for the PR @mitelg :1st_place_medal:

PR #1993, opened 2025-11-03
chore: updates docker install

lasomethingsomething:
Updates the Docker install guide.

### DIFF
--- a/guides/installation/setups/docker.md
+++ b/guides/installation/setups/docker.md
@@ -213,7 +213,7 @@ This command drops you into the container’s terminal (you’ll see the prompt 
 docker compose exec web bin/console cache:clear
 ```
 
-**Tip**: When you’re inside the container, you only need `bin/console …`. If you prefer to run commands from your host machine instead, use the full Docker prefix: `docker compose exec web bin/console cache:clear`. 
+**Tip**: When you’re inside the container, you only need `bin/console …`. If you prefer to run commands from your host machine instead, use the full Docker prefix: `docker compose exec web bin/console cache:clear`.
 
 You’ll use the following Makefile commands later on, when you modify frontend or admin code, or develop plugins that affect the UI:
 
@@ -230,6 +230,7 @@ make watch-admin
 # Start a watcher for Storefront
 make watch-storefront
 ```
+
 These will become part of your everyday development workflow.
 
 ## Verify your installation in the browser (optional)


guides/installation/setups/docker.md
```diff
@@ -8,116 +8,263 @@ nav:
...
 
-::: info
-On macOS we recommend OrbStack, instead of Docker Desktop. OrbStack is a lightweight and fast alternative to Docker Desktop, and it is free for personal use. You can follow the official [OrbStack quickstart guide](https://docs.orbstack.dev/quick-start) to install OrbStack.
-:::
+- [Docker](https://docs.docker.com/get-docker/) or [OrbStack](https://docs.orbstack.dev/quick-start) (macOS) installed and running. OrbStack is a fast, free alternative to Docker.
+- make installed on your machine (`apt install make` on Ubuntu, `brew install make` on macOS) 
+- Docker Compose installed on your machine. Docker Desktop provides it automatically. If you're using OrbStack or something else, you can follow the official [Docker Compose installation guide](https://docs.docker.com/compose/install/).
+- Enough disk and network capacity to pull images (~500MB+ per image depending on tags)
+
+## Pre-pull the image (recommended)
```
keulinho:
```suggestion
## Pre-pull the image (optional)
```

I don't think we should recommend that, docker handles that automatically there is no need for this explicit command, 

the tradeoff we are making here is either run this additional command (that is technically not required) or get two additional lines of output when you run the next command -> i would value the simplicity (in terms of # of commands i need to execute to have it running) over not having the cli output in the docker exec command

guides/installation/setups/docker.md
```diff
@@ -8,116 +8,263 @@ nav:
...
+💡 **Tip:** The access key is not for logging in to the Admin. It’s for programmatic access to your storefront’s data via the Store API.
+</details>
+
+If you want to stop the setup, run `make stop`. 
+
+To start it again, use `make up`. 
+
+If you want to remove the containers, run `make down`. This removes all containers but **keeps your data**. 
+
+To remove both containers *and* data, run `docker compose down -v`. You can then do a full reset.
```
keulinho: `make down` is executing a docker compose down internally, so you can execute `make down` in this case as well 

and exactly how to do a full reset is missing, we should either give that information or skip that part

guides/installation/setups/docker.md
```diff
@@ -8,116 +8,276 @@ nav:
...
+
+This command removes all containers and associated networks.
+
+If you also want to remove all data and volumes (for example, to perform a full reset of your environment), run:
+
+```bash
+docker compose down -v
+```
+
+This will delete the containers, networks, and volumes, meaning all stored data will be lost.
```
keulinho:
```suggestion
The `-v` flag will delete the containers, networks, and volumes, meaning all stored data will be lost.
```

that might prevent me and others from thinking `-v` just stands for verbose output

keulinho (approved):

PR #2002, opened 2025-11-07
chore: remove KeyDB

tinect:
KeyDB seems dead.

related to https://github.com/shopware/shopware/pull/13409


Isengo1989 (approved):
Thanks for the PR @tinect :tada:

PR #2009, opened 2025-11-18
feat: switch to yaml config format

shyim:
XML is deprecated within Symfony. Document the way using YAML files


Isengo1989:
@shyim should this already be merged? Since it was introduced in 7.4 and we are still on 7.3

https://symfony.com/blog/new-in-symfony-7-4-deprecated-xml-configuration

shyim:
When we're gonna rewrite. I would also rewrite entirely to Autowire and just mention the config to be enable Autowire. Thats the modern Symfony way and that what we should teach people

PR #2021, opened 2025-11-26
chore: update contribution guide with new release info process

keulinho:
see https://github.com/shopware/shopware/pull/13710


Isengo1989 (approved):
Thank you for the PR @keulinho 🎉 

General question after seeing this. Will the change in the **RELEASE** file also be connected in some way other than in the **blame** history?

Maybe we can add this info for beginners in shopware/shopware or here - when looking at those notes, I want to see what was affected via code sometimes. With the changelog you had it connected visibly via the ISSUE number

Image

PR #2033, opened 2025-12-10
fix: flow reference events

jozsefdamokos:
The list of available events needed updating + removed the action column as it does not make sense

### DIFF
--- a/resources/references/core-reference/flow-reference.md
+++ b/resources/references/core-reference/flow-reference.md
@@ -11,91 +11,91 @@ nav:
   This functionality is available starting with Shopware 6.4.6.0
 :::
 
-| Event | Description |
-|-------|-------------|
-| checkout.customer.before.login | Triggers as soon as a customer logs in |
-| checkout.customer.deleted | Triggers if a customer gets deleted |
-| checkout.customer.double_opt_in_guest_order | Triggers as soon as double opt-in is accepted in a guest order |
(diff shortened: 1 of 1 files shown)


Isengo1989:
> LGTM 👍 . Small nitpick: It might still be worth to document all the available actions we provide out of the box / with SwagCommercial license somewhere else then.
> 
> And depending on the trigger / context, not all of them are available for a given trigger. But I agree that column can get quite confusing

As far as I understand it just tells the user what actions are triggered in the certain flow.  Is it not of any use for developers? Where would you see the new list?

Isengo1989:
@jozsefdamokos I reformatted the table to be better readable in the IDE

ennasus4sun (approved):

MalteJanz:
@Isengo1989 

> As far as I understand it just tells the user what actions are triggered in the certain flow. 

That's not how the flow builder works. The user is able to build their own Flow in the admin UI, where he can choose a single trigger, and depending on the trigger add actions. Not all actions are available for all triggers.

> Is it not of any use for developers? Where would you see the new list?

Good point, because this are the developer docs I don't see much point in listing all the actions here again. We also have user docs about this feature with a similar list (might need to be updated as well?)
https://docs.shopware.com/en/shopware-6-en/settings/Flow-Builder#trigger

Not even sure if the flow trigger list is that interesting for developers, when we have the same in the user docs

Isengo1989 (approved):
Thx for the PR @jozsefdamokos 🎉

PR #2034, opened 2025-12-10
fix: custom document type

jozsefdamokos:
Fixed some errors + some helpful additions in custom document type page

### DIFF
--- a/guides/plugins/plugins/checkout/document/add-custom-document-type.md
+++ b/guides/plugins/plugins/checkout/document/add-custom-document-type.md
@@ -355,7 +355,7 @@ The service definition for our custom renderer would look like this:
 
 Depending on the file type we either get the content with `$this->fileRendererRegistry->render()` or we need to create the content on our own.
 `DocumentFileRendererRegistry` acts as a central registry for document file renderers based on file extensions (e.g., .pdf, .html). It delegates the rendering of documents to the appropriate renderer implementation.
-Therefore, for each registered service that extends the `AbstractDocumentTypeRenderer`, the content of the document can be generated. New types of `AbstractDocumentTypeRenderer` services can be added with `document.renderer` as the tag name and the file extension as a key.
+Therefore, for each registered service that extends the `AbstractDocumentTypeRenderer`, the content of the document can be generated. New types of `AbstractDocumentTypeRenderer` services can be added with `document_type.renderer` as the tag name and the file extension as a key.
 
 ```xml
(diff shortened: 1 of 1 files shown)


guides/plugins/plugins/checkout/document/add-custom-document-type.md
```diff
@@ -341,12 +341,24 @@ Here's what the function does:
 
 Depending on the file type we either get the content with `$this->fileRendererRegistry->render()` or we need to create the content on our own.
 `DocumentFileRendererRegistry` acts as a central registry for document file renderers based on file extensions (e.g., .pdf, .html). It delegates the rendering of documents to the appropriate renderer implementation.
-Therefore, for each registered service that extends the `AbstractDocumentTypeRenderer`, the content of the document can be generated. New types of `AbstractDocumentTypeRenderer` services can be added with `document_type.renderer` as the tag name and the file extension as a key.
+Therefore, for each registered service that extends the `AbstractDocumentTypeRenderer`, the content of the document can be generated. New types of `AbstractDocumentTypeRenderer` services can be added with `document.renderer` as the tag name and the file extension as a key.
```
ennasus4sun: hm, I think this adjustment is not correct...

- The document renderers: invoice, credit note, delivery and so on which do not have this key: https://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/DependencyInjection/document.xml#L60-L111
- The document file type renderes (pdf, html) which have that key: http://github.com/shopware/shopware/blob/trunk/src/Core/Checkout/DependencyInjection/document.xml#L153-L163

so `document_type.renderer` was fine I think.

ennasus4sun:
one more thing:
I had a look into the other existing page right now ([Add Custom Document](https://developer.shopware.com/docs/guides/plugins/plugins/checkout/document/add-custom-document.html)). I did not test it, but I think it is not adding a new document. Looks more like adding a custom config (in the example for delivery-notes) for a specific sales-channel 🤔 (I assume that's the same we can do in the admin already).
But ... it sounds like we create a new document type.
Maybe this can be written more clear by using the word "document config" instead document.

AND:
maybe also our real new document (DocumentRenderer) will need a new config after we added it (just an assumption)

guides/plugins/plugins/checkout/document/add-custom-document-type.md
jozsefdamokos: Ah yes, I was confused about the two concepts too. Should be correct now

PR #2035, opened 2025-12-10
Update index.md

Swvla:
We want to prohibit advertising on the images in the store.


Isengo1989 (approved):
Thx for the PR @Swvla 🎉

PR #2045, opened 2025-12-17
feat: Allow disable search language analyzer via env variable

tamvt:
Documenting the feature from https://github.com/shopware/shopware/pull/13940


Isengo1989 (approved):
Thx for the PR @tamvt 👍

PR #2053, opened 2025-12-19
fix: remove deprecated project config sync

shyim:


shyim:
we could redirect to https://developer.shopware.com/docs/resources/tooling/fixture-bundle/#fixture-bundle

Isengo1989:
> we could redirect to https://developer.shopware.com/docs/resources/tooling/fixture-bundle/#fixture-bundle

@copilot add the redirect for the delted file. You'll find the instructions in the AGENTS.md

PR #2058, opened 2026-01-05
Fix section=>block in CMS block documentation

nickygerritsen:
This showed `section` in the path instead of `block`.


guides/plugins/plugins/content/cms/add-cms-block.md
```diff
@@ -159,7 +159,7 @@ The Storefront template defines how your element appears on the actual storefron
 * **Technical name**: `image-text-reversed` (The `name` property in Step 2)
 * **Extension**: `.html.twig`
 
-**Shopware is expecting the prefix as part of the full filename in `src/Storefront/Resources/views/storefront/section/cms-section-block-container.html.twig`.**
+**Shopware is expecting the prefix as part of the full filename in `src/Storefront/Resources/views/storefront/block/cms-block-container.html.twig`.**
```
bschulzebaek: Actually this path should be correct: it refers to the place where blocks are loaded to provide a bit more technical context. We could rephrase it a bit:

```markdown
[!NOTE]
Block components are loaded in `src/Storefront/Resources/views/storefront/section/cms-section-block-container.html.twig`. This loader expect the following format for block components: `cms-block-image-text-reversed.html.twig`.
```

PR #2068, opened 2026-01-09
add listing routes invalidation info

h1k3r:
StoreAPI routes that return lists do not use return entities for building xtags (because of reverse proxies headers size limits, potential big number of invalidations, etc). This PR adds short note about it and recommended way of working with it.


concepts/framework/http_cache.md
```diff
@@ -94,6 +94,8 @@ For more information about Store API cache invalidation, you can refer to the [C
 
 This is because all data loaded in a Storefront controller is loaded in the core via the corresponding Store API routes and provided with corresponding cache tags. So the tags of the HTTP cache entries we have in the core consist of the sum of all Store API tags generated or loaded during the request. Therefore, the invalidation of a controller route is controlled over the Store API cache invalidation.
 
+For performance and stability reasons, list-type routes are not tagged with all entities returned in the response. This applies to Store API endpoints such as product listings or search results, category listings, or SEO URL listings. These routes instead rely on their TTL as defined by the active HTTP Caching Policy. If the default TTLs are not appropriate for your use case, configure a [custom caching policy](../../guides/hosting/performance/caches.md#fine-tuning-per-route-or-app-hook) for those routes.
```
keulinho: let's add a bit more context here of what the exact reasons are, something like 

```
While it is technically possible to figure out when a entity that is included in a listing respones did change, this will already incur huge performance and resource needs costs. And still then it is impossible to always invalidate the cached pages, because it is practical impossible to figure out when a change to an entity will lead to the case that that entity needs to be included in the cached responses where it currently is not concluded. Therefore with enabled caching it is basically impossible to guarantee immediate consistency of the cached data, so made the tradeoff in favor of performance and resource efficiency here
```

PR #2084, opened 2026-01-27
feat: remove old config

shyim:


Isengo1989:
@lernhart you want to update the used software versions? 

For now I would say lets close the PR as gitbook is still used and as it seems devenv as well.

PR #2114, opened 2026-02-11
feat: add required ipv6 records for apex domains

tombojer:


products/paas/shopware/cdn/index.md
```diff
@@ -23,32 +23,63 @@ Fastly serves as the primary CDN solution for Shopware PaaS Native, delivering e
...
-Configure a `A` with your custom domain's DNS to point to:
+```dns
+shop.example.com.  IN  CNAME  cdn.shopware.shop.
+```
+
+#### For Apex Domains
+
+If your custom domain is an apex/root domain (e.g., `example.com`), you need to configure multiple record types:
+
+**1. IPv4 routing** - Create four `A` records:
```
renaudhager: I think four needs to be removed here.

drzombey (approved):

products/paas/shopware/cdn/index.md
tombojer: It is important to set all four records otherwise the validation will fail.

I can change it to `all A records` if you want.

renaudhager:
Can we merge this @Isengo1989 ?

PR #2118, opened 2026-02-13
Add good practice guides

keulinho:
First draft to collect good practices for external devs to follow 

mostly intended to give guidance and link out to the correct existing section in the docs


guides/good-practices/upgrades.md
```diff
@@ -0,0 +1,40 @@
...
+  title: Upgrades
+  position: 4
+
+---
+
+# Upgrades
+
+Upgrade plans should protect customer data and keep extensions compatible. Rehearse upgrades on staging with production-like data before touching production, and keep recovery steps ready.
+
+## Cross-cutting practices
```
RigaPaul: It's also worth mentioning both keeping up with updates (ie. update once a quarter) and also to make use of the backwards compatability that is added to new features - these allow a cross-over period when new feature is available but behind a feature flag -> these times should be used to do the upgrade without it being on the critical path.

guides/good-practices/deployments.md
```diff
@@ -0,0 +1,33 @@
...
+---
+
+# Deployments
+
+Deploy with the same artifacts you built in CI, keep configuration outside the codebase, and make database changes predictable. Separate build from runtime wherever possible so releases stay repeatable and reversible.
+
+## Cross-cutting practices
+
+- Roll forward by default; keep rollbacks minimal (DB-aware and version-pinned) and practice them.
+- Use maintenance mode for schema-changing releases; add health checks and smoke tests post-deploy before exiting maintenance.
```
shyim: mention those system:check thingy?

PR #2138, opened 2026-02-25
Fix redirects

sushmangupta:


Isengo1989 (changes requested):
setups should be changed to

guides/installation/setups/ instead of html, because we also use that syntax in the URL and therefor this is what people share / link to. Comparison on other verions help for that

Image

Gotta check if this still works then for the finer redirects e.g. **guides/installation/setups/symfony-cli.html: guides/installation/legacy-setups/symfony-cli-setup.html** Might could be interpreted as folder redirect. Check on stage after the changes -> *shopware-frontends.vercel.app

PR #2198, opened 2026-03-19
feat: Add telemetry documentation for Shopware tools

shyim:


lasomethingsomething (approved):
+1 for this idea and effort. 
- Might consider making it explicit earlier that this is a technical transparency notice, not a privacy notice. You might retitle the file to represent that: "transparency-notice.md."
- would have legal look at this (not a blocker for publishing) to confirm we've not put out/projected guarantees we can't uphold somehow. I've softened language in places with that in mind.

PR #2208, opened 2026-03-27
Create not allowed features in the store

Swvla:
We need a new separate page that lists which features are allowed and which are not.


sushmangupta:
This info is added in https://github.com/shopware/docs/pull/2270

PR #2212, opened 2026-03-30
Revert "Add documentation for Storefront Components"

tobiasberge:
Reverts shopware/docs#2182

We'd like to to revert the docs for the component system for now.
With [this change](https://github.com/shopware/shopware/pull/15820) vite will be introduced as a bundler to compile component JS. It also allows adding external dependencies (node_modules) more easily for 3rd party developers.

This will result in a breaking change in the currently documented way of adding component JavaScript in the theme.json. We want to finish the JS part of the component system first. We can bring back the docs afterwards with the necessary changes. In the meantime we don't want to actively promote the current solution because it will change later.


mitelg (approved):

Isengo1989 (approved):
Thx for the PR  @tobiasberge 👍

PR #2213, opened 2026-03-30
feat: add more info from questions raised by customers

TrayserCassa:


renaudhager (approved):

Isengo1989 (approved):
Thanks for the PR @TrayserCassa 👍 


Small hint, best don´t use .**html** in the link ending. IDEs will better recognize wrong links without, and we handle the links in VitePress also without.

Image

PR #2223, opened 2026-04-03
feat: upgrade ci actions

bojanrajh:
This PR upgrades all actions in CI to their latest version.


Isengo1989 (approved):
LGTM 👍 

I guess we wanna move to pinned versions in the future, right? Specific reason or just following best practice?

PR #2233, opened 2026-04-10
Add entity single select for mail templates sample showing label-property usage

Timothylul:
Add entity single select for mail templates sample showing label-property usage

## Summary

Adds a sample on how to use the label-property field within the config.xml as I could not find any documentation showing how to use this. 

## Related links

https://developer.shopware.com/docs/guides/plugins/plugins/plugin-fundamentals/add-plugin-configuration.html

## Checklist

- [ ] I reviewed affected links, code samples, and cross-references, including `PageRef` references where relevant.
- [ ] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted.
- [ ] I updated `.wordlist.txt` (and sorted it) if spellcheck flags new legitimate terms.
- [ ] This pull request is ready for review.

### DIFF
--- a/guides/plugins/plugins/plugin-fundamentals/add-plugin-configuration.md
+++ b/guides/plugins/plugins/plugin-fundamentals/add-plugin-configuration.md
@@ -271,7 +271,18 @@ Here are some examples:
 </component>
 ```
 
-Stores the ID of the selected product into the system config.
+### Entity single select for mail templates
+
+```html
+<component name="sw-entity-single-select">
+    <name>exampleMailTemplate</name>
+    <entity>mail_template</entity>
+    <label-property>description</label-property>
+    <label>Choose a mail template for the plugin configuration</label>
+</component>
+```
+
+Stores the ID of the selected mail template into the system config. Displays the mail templates description in the administration.
 
 ### Entity multi ID select for products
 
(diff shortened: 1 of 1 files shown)


Isengo1989:
@Timothylul thx for the PR. The docs state that there are 5 possible types, you added a second one for :

\<component name="sw-entity-single-select"\> - there is one for **sw-entity-single-select** already. Anything in particular that you felt was missing instead? Trying to avoid duplicates and grow the list of examples that have the same component type. We can still add it, as it gives extra information for **label-property**, but I am curious if some info was maybe unclear.

Isengo1989:
@Timothylul could you please take a look, thx

Timothylul:
Hi @Isengo1989 
I could not find anything regarding the label-property. Just thought it would be the easiest way to add another example showing how to use this. But as you said the sw-entity-single-select component would be duplicated. Maybe it is better to just add the label-property to the existing example with an explanation

Isengo1989:
> Hi Micha Hobert (@Isengo1989) I could not find anything regarding the label-property. Just thought it would be the easiest way to add another example showing how to use this. But as you said the sw-entity-single-select component would be duplicated. Maybe it is better to just add the label-property to the existing example with an explanation

OK, got it, I misread that this was about the mail_template, but further check showed why this is relevant, also for other entities. The issue here is that if there is no "name" on the entity, it is just blank. Probably the reason why you checked for alternatives in the first place.

Will add that info for others in this PR

Isengo1989 (approved):

PR #2314, opened 2026-05-29
Clarify CLI formatter Docker requirements

shyim:
## Summary
- Explain that formatter, validation, and automatic refactoring commands use PHP and Node.js tooling under the hood
- Clarify that Docker is recommended because the image includes those dependencies
- Add context for the `/ext` and `/project` container mount paths versus local paths
- Replace local validation examples that used `/ext` with `/path/to/your/extension`

## Verification
- `git diff --check`
- Content assertion for expected clarification text
- `make spellcheck` using Docker

Note: Two unrelated untracked talk files are present locally and were temporarily moved out before running the Docker spellcheck, then restored.


Isengo1989 (approved):
Thx for the PR @shyim 👍

PR #2338, opened 2026-06-17
fix: remove wrong parameter

TrayserCassa:
In the latest release we have removed this parameter, this is a left over. Should be a quick win


Isengo1989 (approved):
Thx for the fix @TrayserCassa 👍

PR #2342, opened 2026-06-18
docs(elasticsearch): document search relevance configuration options

vienthuong:
## Summary

Documents the configurable search-relevance options introduced with the recent Elasticsearch product-search improvements.

- **Environment variables** — adds `SHOPWARE_ES_DIMENSION_NORMALIZE` to the reference table. It normalizes dimensional notation (for example `5 x 70` → `5x70`) in the technical-term analyzer chains, is disabled by default, and requires a full reindex (`bin/console es:index`) to take effect.
- **Advanced Search → Modify Search Logic** — documents strictness-based matching (percentage `minimum_should_match`) as an alternative to the binary `AND` / `OR` search behavior, including how the `0%` / `50%` / `100%` levels behave. The presets are opt-in; `AND` / `OR` remains the default.
- **Advanced Search → Language analyzers** — documents German compound-word decomposition: a `dictionary_decompounder` filter that splits compound nouns at index time using editable, per-language dictionaries (`advanced_search_compound_dictionary` / `advanced_search_stopword_dictionary`), seeded with a default German root-word list.


tamvt (approved):

guides/plugins/plugins/integrations/elasticsearch/index.md
```diff
@@ -8,3 +8,48 @@ nav:
...
+
+The `elasticsearch.search` section holds further settings you can override in `config/packages/elasticsearch.yaml`:
+
+```yaml
+elasticsearch:
+    search:
+        timeout: 5s                     # request timeout per search
+        term_max_length: 300            # search terms longer than this are truncated
+        search_type: query_then_fetch   # OpenSearch search type
+        # precision_threshold: 40000    # cardinality accuracy for grouped product counts (higher = more memory)
```
Isengo1989, line 46: @vienthuong is the comment on the value intentional since it should not be used?

# precision_threshold: 40000

vienthuong: hi @Isengo1989, the value is intentional as an example value which is the default value anyway but we don't want to send it on indexing unless they explictly set it because they might already modified the value on the opensearch server on their own.

Isengo1989 (approved):
Thx for the PR @vienthuong 💯

PR #2343, opened 2026-06-18
docs: clearify on token usage

TrayserCassa:
Adding more warning and caution to personal access tokens. And for ci/cd use service accounts


Isengo1989 (approved):
Thx for the PR @TrayserCassa

PR #2349, opened 2026-06-24
docs: document DynamicallyScheduledTaskHandler for custom rescheduling

shyim:
## Summary

Documents the new `DynamicallyScheduledTaskHandler` interface in the "Add Scheduled Task" guide. A handler implementing it can compute its own next execution time (e.g. the timestamp of the next pending record) instead of the default `now + runInterval` schedule, while the executor still persists the schedule.

Adds a new "Dynamic rescheduling" section with an `Available since Shopware version 6.7.13.0` note and a full handler example.

## Related links

- Platform PR: shopware/shopware#17293

## Checklist

- [ ] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted.
- [x] Any required dependent changes in downstream modules have already been merged and published.

## Notes

New terms are all wrapped in code fences/backticks, so no `.wordlist.txt` changes were needed — `make spellcheck` and the markdown style check both pass locally. The "available since" version (6.7.13.0) matches the upcoming release the platform PR targets; adjust if the platform PR lands in a different version.

### DIFF
--- a/guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
+++ b/guides/plugins/plugins/plugin-fundamentals/add-scheduled-task.md
@@ -116,7 +116,7 @@ Every five minutes, Shopware will dispatch the task to the message bus and the h
 Available since Shopware version 6.7.13.0
 :::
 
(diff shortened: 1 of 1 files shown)


shyim:
I added blocked label, as we wait until it's released right?

Isengo1989:
> I added blocked label, as we wait until it's released right?

This always depends on the PR; ideally, we merge when it is released. Sometimes we want to give external developers a longer "heads up".

As this is only one new interface and the version is clearly stated beforehand, I would go with "merge now".

Isengo1989 (approved):
Thx for the PR @shyim 👍

PR #2361, opened 2026-07-01
docs(hosting): document keeping the Docker setup up to date

shyim:
## Why

User feedback asked us to document the recommended way to keep a Docker-based Shopware environment up to date — and specifically the combined upgrade path (e.g. Docker image with PHP 8.2 / Shopware 6.6 → PHP 8.5 / Shopware 6.7). The existing [Docker Image guide](https://developer.shopware.com/docs/guides/hosting/installation-updates/docker.html) helps users get a container running but never explains how to maintain it afterwards, which leaves shops (especially smaller ones) running outdated, unpatched PHP.

## What

Adds a **"Keeping the setup up to date"** section to `guides/hosting/installation-updates/docker.md`:

- Clarifies the **three separate things** that get updated (PHP base image, Shopware, own code) and how each differs — directly answering the "what up to date? php? shopware? plugins?" question from the thread.
- An **"Am I affected?"** check: `php -v` inside the container plus image scanning with **Trivy / Grype**.
- Recommends **scanning on a schedule and rebuilding only when a scan finds a vulnerability**, rather than blind daily rebuilds.
- The full **rebuild-and-redeploy** commands tied to the compose setup already on the page, with an explicit warning that *build ≠ deploy*.
- The recommended **combined Shopware + PHP major upgrade path** as separate, ordered steps, with backups and a warning against multi-major jumps.

Written for the non-expert shop owner, since that's who most often hits this problem.

## Notes

- Cross-links to the existing `performing-updates.md` and `system-requirements.md` rather than duplicating the general update procedure.
- Verified `::: warning`/`::: danger` containers and `<PageRef>` (no `.md` extension) match existing repo conventions.

### DIFF
--- a/guides/hosting/installation-updates/docker.md
+++ b/guides/hosting/installation-updates/docker.md
@@ -249,7 +249,7 @@ If you are not sure how old your running shop is, check the PHP version inside t
 docker compose exec web php -v
 ```
 
-Compare that with the [latest PHP releases](https://www.php.net/supported-versions.php). If the patch version (the last number, e.g. the `12` in `8.3.12`) is old, your container has not been rebuilt in a while and is missing security fixes. That is the signal to do the rebuild below.
+Compare that with the [latest PHP releases](https://www.php.net/supported-versions.php). If a newer patch version is available (the last number, e.g. a higher number than the `12` in `8.3.12`), your container has not been rebuilt in a while and is missing security fixes. That is the signal to do the rebuild below.
 
 For the full picture — not just PHP, but every package in the image — scan the image with a vulnerability scanner such as [Trivy](https://trivy.dev/) or [Grype](https://github.com/anchore/grype). Trivy needs no setup and reports all known vulnerabilities (CVEs) in the image:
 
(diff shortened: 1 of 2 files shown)


guides/hosting/installation-updates/docker.md
```diff
@@ -225,6 +225,108 @@ volumes:
...
+
+### Am I affected? Check what you are running
+
+If you are not sure how old your running shop is, check the PHP version inside the running container:
+
+```bash
+docker compose exec web php -v
+```
+
+Compare that with the [latest PHP releases](https://www.php.net/supported-versions.php). If the patch version (the last number, e.g. the `12` in `8.3.12`) is old, your container has not been rebuilt in a while and is missing security fixes. That is the signal to do the rebuild below.
```
jenskueper: is old -> if there is a newer version available

renaudhager (approved):
LGTM.

sushmangupta (approved):
Thanks for teh PR @shyim :)

Isengo1989 (approved):

PR #2370, opened 2026-07-07
chore(docs): document rate limiter defaults

BrocksiNet:
## What

Extends the Rate Limiter hosting guide with a complete overview of the rate limiters Shopware ships by default:

- New **Default rate limiters** table listing every core limiter with its `Since` version, policy, reset window, and limits.
- Clarifies in the Overview that these limiters are **always active** and that `shopware.yml` is only used to *override* the defaults, not to enable them.
- Reworks the Configuration section to make clear that the `rate_limiter` map is merged into the defaults, so you only configure the values you want to change.
- Removes the separate version info block, since the introduction version now lives in the table's `Since` column.

## Why

The page previously listed only 6 of the default limiters by name and never stated their actual values. This led to recurring questions about whether the rate limiter works at all without adding `shopware.api.rate_limiter` to `shopware.yml` (it does, the core ships the defaults). Documenting the concrete values and the "always active" behaviour answers this directly.

## Source of truth

Values were taken from `src/Core/Framework/Resources/config/packages/shopware.yaml`. Each `Since` version was verified against the release folder its changelog entry landed in (for older limiters) or the release tag the introducing commit first shipped in (for recent ones).

The table also notes that values can change between versions and points readers at the config file, and asks contributors to add a row whenever a new default limiter is introduced.

### DIFF
--- a/guides/hosting/infrastructure/rate-limiter.md
+++ b/guides/hosting/infrastructure/rate-limiter.md
@@ -15,6 +15,35 @@ This functionality is available starting with Shopware 6.4.6.0.
 
 Shopware 6 provides certain rate limits by default that reduces the risk of brute-force attacks for pages like login or password reset.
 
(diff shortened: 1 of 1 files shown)


Isengo1989 (approved):
Thanks for the PR @BrocksiNet 👍

PR #2414, opened 2026-07-31
docs: document new config for disabling path cache busting for media

bneumann97:
## Summary

Media replacement currently changes both the physical path and the ts query parameter. URLs retained in cached pages, emails, or search indexes can therefore point to a deleted object and return 404.

Adds the opt-in shopware.cdn.path_cache_buster configuration, defaulting to true.
When disabled, media and thumbnail paths remain stable on replacement while the existing ?ts= query parameter continues to provide cache busting. This applies to all built-in media path strategies, including physical_filename.

## Related links

https://github.com/shopware/shopware/pull/18860

## Checklist

- [x] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted.
- [x] Any required dependent changes in downstream modules have already been merged and published.


guides/hosting/infrastructure/filesystem.md
```diff
@@ -374,6 +374,23 @@ SHOPWARE_CDN_STRATEGY_DEFAULT=id
...
+By default, Shopware adds the upload timestamp to the physical media path and adds the media update timestamp as a `ts` query parameter to its public URL. You can disable the path timestamp and use query-string cache busting only:
+
+```yaml
+# <project root>/config/packages/prod/shopware.yml
+shopware:
+  cdn:
+    path_cache_buster: false
+```
+
+With this setting, media uploaded or replaced after the configuration change keeps the same physical path. Its `ts` query parameter changes when the media changes, so newly rendered pages request the new version while older URLs continue to resolve.
```
sushmangupta, line 388:
```suggestion
With this setting, media uploaded or replaced after the configuration change keeps the same physical path. Its `ts` query parameter updates when the media changes, so newly rendered pages request the new version while older URLs continue to resolve to the previous version.
```

sushmangupta (changes requested):

guides/hosting/infrastructure/filesystem.md
```diff
@@ -374,6 +374,23 @@ SHOPWARE_CDN_STRATEGY_DEFAULT=id
 Changing the strategy on a shop with existing media changes the generated paths for those files, which will break existing URLs unless the files are moved to match. Choose a strategy before going live, or plan a migration of the affected paths.
 :::
 
+### Media path cache busting
+
+By default, Shopware adds the upload timestamp to the physical media path and adds the media update timestamp as a `ts` query parameter to its public URL. You can disable the path timestamp and use query-string cache busting only:
```
sushmangupta:
```suggestion
By default, Shopware appends the upload timestamp to the physical media path and includes the media update timestamp as a `ts` query parameter in its public URL. You can disable the path timestamp and use query-string cache busting only:
```

sushmangupta (changes requested):

guides/hosting/infrastructure/filesystem.md
bneumann97: The older URL's don't resolve to the previous version. If you replace an image, the URL stays the same but it will show the new version of the image. That's for SEO, cached pages and mailing reasons

sushmangupta (approved):

PR #2467, opened 2026-08-21
docs(hosting): explain SHOPWARE_ADMIN_ES_REFRESH_INDICES and align its default

vienthuong:
## Summary

`SHOPWARE_ADMIN_ES_REFRESH_INDICES` is documented in two places, and neither says what it does.

- **Environment variable reference** described it as "Refresh administration indices" — a restatement of the name that reads like Elasticsearch's `_refresh` API. What it actually controls is `AdminSearchRegistry::refreshIndices()`: on every write of an indexed entity, indexers whose alias is missing get a fresh index + alias, and their `admin_elasticsearch_index_task` row is re-inserted. Replaced with a description of the behaviour, its cost, and its caveat, following the format already used by the `SHOPWARE_ADMIN_ES_THROW_EXCEPTION` row.
- **Admin Elasticsearch setup snippet** set it to `1`. That is the only line in the block diverging from the shipped default — `SHOPWARE_ADMIN_ES_ENABLED=1` is the switch you must flip, and `INDEX_PREFIX` / `INDEXING_BATCH_SIZE` / `THROW_EXCEPTION` all match their defaults. So a copy-paste opted readers into an `existsAlias` call per indexer (17 in core) on every admin write, plus the possibility of an empty index behind a live alias, without a word of explanation — while `bin/console es:admin:index`, the thing that makes it unnecessary, is documented three lines below. Set to `0` and added an info block on when `1` is the right call.

## Related links

- Behaviour: [`AdminSearchRegistry::refresh()` / `refreshIndices()`](https://github.com/shopware/shopware/blob/trunk/src/Elasticsearch/Admin/AdminSearchRegistry.php)
- Shipped default `""`: [`elasticsearch.yaml`](https://github.com/shopware/shopware/blob/trunk/src/Elasticsearch/Resources/config/packages/elasticsearch.yaml)
- Installer default `0`: [`EnvConfigWriter`](https://github.com/shopware/shopware/blob/trunk/src/Core/Installer/Configuration/EnvConfigWriter.php), [`SystemSetupCommand`](https://github.com/shopware/shopware/blob/trunk/src/Core/Maintenance/System/Command/SystemSetupCommand.php)

## Checklist

- [x] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted. — n/a, no pages moved
- [x] I updated `.wordlist.txt` (and sorted it) if spellcheck flags new legitimate terms. — n/a, "indexer"/"indexers" already appear in checked prose (`flags-reference.md`, `extensibility-events-messages.md`)
- [ ] This pull request is ready for review.

## Notes

**One open question for a reviewer who knows the history.** I could find no recorded reason for `=1` in the setup guide — no `->info()` on the config node, no comment on `refreshIndices()`, and the introducing commit (`105501782f8`, NEXT-21467) shipped no changelog line about the flag. I read `=1` as an oversight given the rest of the block, but if it was deliberate — e.g. for managed hosting where nobody runs CLI after installing an extension — then the value should stay `1` and only the info block should land. Happy to drop the value change and keep the prose.

Left as a draft pending that confirmation.


fschmtt (approved):

Isengo1989 (approved):
Thx for the PR @vienthuong 👍

PR #2479, opened 2026-08-25
docs(storefront): use ProductListingCollectSortingEvent for runtime sortings

vienthuong:
## Summary

The "Create individual sorting at runtime" recipe on the [custom sorting guide](https://developer.shopware.com/docs/guides/plugins/plugins/storefront/add-custom-sorting-product-listing.html#create-individual-sorting-at-runtime) does not work and has not worked since v6.6.

It tells you to add a `ProductSortingEntity` to the `sortings` criteria extension from `ProductListingCriteriaEvent`, subscribing with `priority => 500` "to add your sorting before the default shopware logic applies". That was accurate while the sorting logic lived in `ProductListingFeaturesSubscriber`, a listener on the same event. Since shopware/shopware@588c40d51c5 removed that subscriber, `SortingListingProcessor::prepare()` runs before the event is dispatched, so no listener priority can precede it. Following the guide today produces a sorting that appears in the Storefront dropdown but never orders the products — see shopware/shopware#6072.

This updates the section to the new `ProductListingCollectSortingEvent`, which is dispatched before the requested sorting is resolved:

- subscribe to `ProductListingCollectSortingEvent` instead, without the priority workaround and without the stale `ProductSortingNotFoundException` warning
- add the entity via `$event->getSortings()->add(...)` instead of reading and re-writing the `sortings` criteria extension
- drop the now-unused `ProductSortingCollection` import
- note that the event is available since 6.7.15.0, and what the old recipe actually did on earlier versions

## Related links

- Adds the event: shopware/shopware#16213
- Fixes the reported behaviour: shopware/shopware#6072
- Jira: https://shopware.atlassian.net/browse/NEXT-40220

## Checklist

- [x] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted.

## Notes

The dependent-changes box is unchecked because shopware/shopware#16213 is still open. Please merge this only once that lands, so the guide does not document an event that does not exist yet.

No page was moved or renamed, so no redirect is needed. No wordlist entry is needed either: the new identifiers appear only inside code spans and code blocks, the same way `ProductListingCriteriaEvent` did before.

Verified locally with `avtodev/markdown-lint:v1.5` against `markdown-style-config.yml` — no findings.

### DIFF
--- a/guides/plugins/plugins/storefront/howto/add-custom-sorting-product-listing.md
+++ b/guides/plugins/plugins/storefront/howto/add-custom-sorting-product-listing.md
@@ -101,7 +101,7 @@ This event is available since Shopware 6.7.15.0.
 While possible, it is not recommended adding an individual sorting at runtime. If you just wish for your individual sorting to be not editable by users in the Administration, create a migration and set the parameter `locked` to be `true`.
 :::
 
-Here's an example how your subscriber could look like:
+Here's an example of what your subscriber could look like:
 
 ```php
 // <plugin root>/src/Subscriber/ExampleListingSubscriber.php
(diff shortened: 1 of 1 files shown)


Isengo1989 (approved):
Thank you for the PR @vienthuong 👍

PR #1466, opened 2024-09-02
Update folder name to underscore

SirPhoeniix:
I have noticed that the documentation says that the folder for your translations should be de-DE or en-GB. I have updated this in this pull request because I had to name the folders de_DE and en_GB to get translations working.

PR #2487, opened 2026-08-31
docs: document `postinstall` copy-out workaround for shopware-cli #1466

shyim:
Documents the known workaround from [shopware/shopware-cli#1466](https://github.com/shopware/shopware-cli/issues/1466) so theme authors can keep using npm packages (e.g. `@fortawesome/fontawesome-free`) from SCSS while the upstream fix is still open.

### Problem

`shopware-cli project storefront-build` fails for a pure-SCSS theme that imports from `node_modules` because:

* it only runs `npm install` for storefront extensions with a JS entry point, and
* even with a `main.js`, the storefront-root `node_modules` is removed before `theme:compile` runs.

### Changes

* New section **"Using npm packages in a pure-SCSS theme (no JS entry point)"** in `guides/plugins/plugins/dependencies/using-npm-dependencies.md` describing the `postinstall` copy-out workaround: empty `main.js`, `package.json` `postinstall` that copies SCSS into `.vendor/node_modules/...` and webfonts into `public/static/fonts/`, updated `theme.json`/`SCSS` paths, and `.gitignore` entries.
* Short pointer with anchor link from `guides/plugins/themes/styling/add-css-js-to-theme.md` so theme authors find the workaround from the theme styling guide.
* Added `webfonts` to `.wordlist.txt`.

### Quality

* `make spellcheck` passes.
* `make fix` (markdown lint) passes.

Refs shopware/shopware-cli#1466.

### DIFF
--- a/guides/plugins/plugins/dependencies/using-npm-dependencies.md
+++ b/guides/plugins/plugins/dependencies/using-npm-dependencies.md
@@ -126,6 +126,93 @@ Build the Storefront using:
 shopware-cli project storefront-build
 ```
 
+## Using npm packages in a pure-SCSS theme (no JS entry point)
+
+If your plugin is a **theme** and only consumes npm packages from SCSS (for example `@fortawesome/fontawesome-free` or any other package referenced via `@import` in your `theme.json` `style` entries), you will run into a chicken-and-egg problem with `shopware-cli project storefront-build`:
+
+* `shopware-cli` only runs `npm install` for a storefront extension when it has a JavaScript entry point (`src/Resources/app/storefront/src/main.js`).
+* Even when a `main.js` is present, `shopware-cli` runs the asset build (webpack) first and **then deletes the storefront-root `node_modules` directory** before `theme:compile` runs.
+
+The result is that a `theme.json`/`SCSS` `@import` like `app/storefront/node_modules/@fortawesome/fontawesome-free/scss/fontawesome` cannot be resolved, and `theme:compile` aborts with:
+
+```text
+Unable to compile the theme "MyTheme". Unable to resolve file
+"Resources/app/storefront/node_modules/@fortawesome/fontawesome-free/scss/fontawesome.scss".
+```
+
+### Workaround: copy required assets out via `postinstall`
+
+Because the `postinstall` script of a storefront extension's `package.json` runs **before** the root `node_modules` directory is deleted, you can copy the parts of the package that `theme:compile` needs into a persistent (git-ignored) folder inside the theme. After that, the SCSS imports and `theme.json` `style` entries must point at that folder instead of `node_modules`.
+
+The nested folder must contain a `node_modules` path segment, because `shopware-cli project format` / `validate` only ignore the storefront-**root** `node_modules` and `shopware-cli` only deletes that one level. A nested copy survives both steps.
+
+#### 1. Provide an empty JS entry point
+
+Add an empty `src/Resources/app/storefront/src/main.js` so `shopware-cli` installs the theme's npm dependencies and runs its lifecycle scripts:
+
+```javascript
+// <plugin root>/src/Resources/app/storefront/src/main.js
+// Intentionally empty. Present so shopware-cli installs this theme's npm
+// dependencies and runs the package.json "postinstall" below.
+```
+
+#### 2. Copy the needed files out of `node_modules` in `postinstall`
+
+```json
+// <plugin root>/src/Resources/app/storefront/package.json
+{
+    "dependencies": {
+        "@fortawesome/fontawesome-free": "^6.1.1"
+    },
+    "scripts": {
+        "postinstall": "rm -rf .vendor && mkdir -p .vendor/node_modules/@fortawesome/fontawesome-free && cp -R node_modules/@fortawesome/fontawesome-free/scss .vendor/node_modules/@fortawesome/fontawesome-free/ && mkdir -p ../../public/static/fonts && cp node_modules/@fortawesome/fontawesome-free/webfonts/fa-* ../../public/static/fonts/"
+    }
+}
+```
+
+The script does three things:
+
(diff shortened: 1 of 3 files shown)


guides/plugins/plugins/dependencies/using-npm-dependencies.md
```diff
@@ -126,6 +126,93 @@ Build the Storefront using:
 shopware-cli project storefront-build
 ```
 
+## Using npm packages in a pure-SCSS theme (no JS entry point)
+
+If your plugin is a **theme** and only consumes npm packages from SCSS (for example `@fortawesome/fontawesome-free` or any other package referenced via `@import` in your `theme.json` `style` entries), you will run into a chicken-and-egg problem with `shopware-cli project storefront-build`:
```
tobiasberge, line 146: Idk I find the chicken-egg wording a bit strange for the docs.

```suggestion
If your plugin is a **theme** and only consumes npm packages from SCSS (for example `@fortawesome/fontawesome-free` or any other package referenced via `@import` in your `theme.json` `style` entries), you will not be able to run `shopware-cli project storefront-build` successfully:
```

guides/plugins/plugins/dependencies/using-npm-dependencies.md
```diff
@@ -126,6 +126,93 @@ Build the Storefront using:
...
+
+* `shopware-cli` only runs `npm install` for a storefront extension when it has a JavaScript entry point (`src/Resources/app/storefront/src/main.js`).
```
tobiasberge, line 148: I would put this part BEHIND "The result is that a `theme.json`/`SCSS` `@import` like `app/storefront/node_modules/@fortawesome/fontawesome-free/scss/fontawesome` cannot be resolved, and `theme:compile` aborts with:".

So the reading is order:
1. Will not working
2. Error message
3. And then the reason (background info) why its not working

guides/plugins/plugins/dependencies/using-npm-dependencies.md
```diff
@@ -126,6 +126,93 @@ Build the Storefront using:
...
+```text
+Unable to compile the theme "MyTheme". Unable to resolve file
+"Resources/app/storefront/node_modules/@fortawesome/fontawesome-free/scss/fontawesome.scss".
+```
+
+### Workaround: copy required assets out via `postinstall`
+
+Because the `postinstall` script of a storefront extension's `package.json` runs **before** the root `node_modules` directory is deleted, you can copy the parts of the package that `theme:compile` needs into a persistent (git-ignored) folder inside the theme. After that, the SCSS imports and `theme.json` `style` entries must point at that folder instead of `node_modules`.
+
+The nested folder must contain a `node_modules` path segment, because `shopware-cli project format` / `validate` only ignore the storefront-**root** `node_modules` and `shopware-cli` only deletes that one level. A nested copy survives both steps.
```
tobiasberge, line 162: Not sure if I understand this part. What does nested folder mean? Maybe provide an example?

guides/plugins/plugins/dependencies/using-npm-dependencies.md
```diff
@@ -126,6 +126,93 @@ Build the Storefront using:
...
+
+### Why this works and what to expect
+
+* `shopware-cli project storefront-build` runs `npm install` for the theme (because `main.js` exists), which triggers `postinstall` and copies the files into `.vendor/` and `public/static/fonts/`.
+* The webpack build runs, the storefront-root `node_modules` is deleted, and then `theme:compile` runs against the persistent `.vendor/node_modules/...` copy. The build exits 0.
+* `shopware-cli project format` / `validate` do not flag the copied files because they live under a nested `node_modules/` path.
+* The webfonts end up under `public/bundles/<theme>/static/fonts/` after compilation, which is what the SCSS expects at runtime.
+
+This workaround is known to be fragile: it depends on `shopware-cli` running npm lifecycle scripts and on the cleanup deleting only the storefront-root `node_modules`. If you do not need npm packages in your theme, prefer keeping the theme free of `node_modules` imports. A longer-term fix (installing npm dependencies for a JS-less theme and deferring the `node_modules` cleanup until after `theme:compile`) is being discussed in [shopware/shopware-cli#1466](https://github.com/shopware/shopware-cli/issues/1466).
```
tobiasberge, line 242: Maybe you can also point to the normal asset adding workflow here as an alternative.

guides/plugins/plugins/dependencies/using-npm-dependencies.md
shyim: Good point — replaced "chicken-and-egg" with "you will not be able to run ... successfully". Fixed in f7daa36b.

guides/plugins/plugins/dependencies/using-npm-dependencies.md
shyim: Done — the error message now comes first, followed by "The reason is that:" and the two bullets explaining the cause. Fixed in f7daa36b.

guides/plugins/plugins/dependencies/using-npm-dependencies.md
```diff
@@ -126,6 +126,93 @@ Build the Storefront using:
...
+    "style": [
+        "app/storefront/.vendor/node_modules/@fortawesome/fontawesome-free/scss/fontawesome.scss",
+        "app/storefront/src/scss/base.scss"
+    ]
+}
+```
+
+```scss
+// <plugin root>/src/Resources/app/storefront/src/scss/base.scss
+@import '../../.vendor/node_modules/@fortawesome/fontawesome-free/scss/fontawesome';
```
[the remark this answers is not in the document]
Isengo1989: @shyim not sure about this, can you check?

Isengo1989 (approved):
@shyim  Left one comment where I am not sure about, rest looks good to me, thanks for the PR 👍 

Just added some code-blocks to avoid inline comments in JSON

guides/plugins/plugins/dependencies/using-npm-dependencies.md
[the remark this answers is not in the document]
shyim: yes thats valid for the referencing .woff files

PR #2500, opened 2026-09-04
Documentation Update: feat(administration): shim for sfc override templates to twig templates (PR #20006)

shopwareBot:
# Documentation Update for PR #20006

This PR updates documentation based on changes from [PR #20006](https://github.com/shopware/shopware/pull/20006) in the source repository.

## 📊 Change Analysis
- **Impact Score**: 5/10
- **Files Updated**: 4 documentation files
- **Qdrant Collection**: `shopware_docs`

## 📝 Documentation Files Updated
- `guides/upgrades-migrations/administration/vue-native.md`
- `guides/upgrades-migrations/administration/index.md`
- `guides/upgrades-migrations/administration/vue-migration-build.md`
- `guides/upgrades-migrations/administration/vite.md`

## 🎯 Key Areas Addressed
- Update `guides/upgrades-migrations/administration/vue-native.md` to document that native Vue SFC Administration overrides can now target many legacy Twig-based core components through a compatibility shim. Explain the new developer-visible behavior change: overrides that previously rendered nothing against Twig components may now work, and clarify that this is partial compatibility during the Twig-to-native migration rather than full parity.
- Revise `guides/upgrades-migrations/administration/vue-native.md` to add explicit limitations and troubleshooting guidance for the new override path. Include that support depends on detected Twig block targets, that some Twig structures are still unsupported, and especially that blocks containing multiple named slot templates are not currently supported and may still require legacy Twig-based override techniques.
- Update `guides/upgrades-migrations/administration/vite.md` in the section about automatic override-component handling to reflect the changed build behavior: override templates are now analyzed at build time for Twig block usage and registered before runtime template resolution. This should be described as part of the Administration extension compilation pipeline so developers understand why mixed Twig/SFC overrides now work in some cases.
- Revise `guides/upgrades-migrations/administration/vue-migration-build.md` to cover this interim bridge behavior in the migration build story. Document that the migration period now supports a hybrid state where core Administration components may still be Twig-based while plugin extensions can already move to native `.vue` override patterns for supported blocks.
- Update `guides/upgrades-migrations/administration/index.md` to surface this as a notable Administration migration topic. Add or adjust navigation/overview text so plugin developers can discover that native override support has expanded to include many legacy Twig-based component targets, with caveats and a pointer to the detailed native Vue migration guidance.

## 📋 Summary
Administration extension behavior changed in a documentation-relevant way: native Vue SFC overrides can now render into many legacy Twig-based core components through a build-time and runtime shim. No API, schema, or deployment docs are affected, but Administration migration and Vite/build guidance should be updated to explain the new capability, its limitations, and the hybrid Twig/SFC migration path for plugin developers.

---
*This documentation was automatically generated based on code change analysis.*


bojanrajh:
@copilot Please fix Markdown issues.
```
guides/upgrades-migrations/administration/index.md:23:188 MD047/single-trailing-newline Files should end with a single newline character
guides/upgrades-migrations/administration/vite.md:268:327 MD047/single-trailing-newline Files should end with a single newline character
guides/upgrades-migrations/administration/vue-migration-build.md:315:310 MD047/single-trailing-newline Files should end with a single newline character
guides/upgrades-migrations/administration/vue-native.md:536:47 MD047/single-trailing-newline Files should end with a single newline character
```

davidtraum (commented):
I think the PR does not need to be documented to this extend. It does not introduce a new user facing api but more of an internal compatibility layer.

I still checked the content. It contains some valid points, but it described a lot more than there is actually in this PR and also made up some information to give more (not always accurate) context. For example the section "When to use this approach" does not really make sense because you would have to look at the whole new native extension system and not just this PR.
Maybe it would be more accurate if the rest of the (currently experimental / not yet advertised) native extension system would already be fully documented and it did not need to add this much context.

We are also already planning to do a full documentation of the native extension system when it is ready and will include a note there that this feature exists.

So this documentation is not really necessary here, but it might work good for other PRs.

PR #2517, opened 2026-09-16
fix(paas-native): adjusted clone, domain and deployment documentation

axasanya:
## Summary

Corrects five factual inaccuracies in the PaaS Native documentation, found during a full review of the product section:

- **Clone Application** — stated that cloning copies the codebase. It does not; it copies database and object storage only, which is why the target application must already exist and should be on the same commit as the source.
- **Applications / DNS** — instructed a `CNAME` to `cdn.shopware.shop`. That is invalid for apex domains and omits the `TXT` ownership challenge. Replaced with a pointer to the CDN page, which carries the complete record set, plus the requirement that records propagate *before* `sw-paas domain create` runs, since the platform validates them in real time.
- **Applications / Deployment behavior** — described deployments as zero downtime unconditionally. That does not hold when the deployment includes database migrations, so the claim is now scoped and the failure mode stated.
- **OpenSearch** — the `when@dev` web profiler configuration sat in a trailing `## Note`, but skipping it fails the deployment. Moved into a `## Before you start` warning. Also added that the deployment reports success before the index is built.
- **Quickstart** — Step 3 ran `sw-paas application deploy create` after `sw-paas application create`, which already builds and deploys. Replaced with a verification step and `sw-paas open storefront`.

## Checklist

- [x] I added or updated redirects in `.gitbook.yaml` if pages were moved, renamed, or deleted. — n/a, no pages moved or renamed
- [x] Any required dependent changes in downstream modules have already been merged and published. — n/a

## Notes

Content corrections only — no pages added, moved or removed. Each change is verifiable against current CLI and platform behaviour. Happy to split this into per-page commits if that reviews more easily.


sushmangupta:
@axasanya , please check and resolve copilot suggestions

